PR #29, opened 2022-06-10
Allow entering path as argument to the init

shamsartem:


src/commands/init.ts
```diff
@@ -47,72 +46,34 @@ import { getArtifactsPath } from "../lib/pathsGetters/getArtifactsPath";
...
+  ];
   static override usage: string = usage(this);
   async run(): Promise<void> {
     const { args, flags } = await this.parse(Init);
     const isInteractive = getIsInteractive(flags);
-    const projectName: unknown = args[NAME_ARG];
-    assert(projectName === undefined || typeof projectName === "string");
-    await init(this, isInteractive, projectName);
+    const nameOrPath: unknown = args[NAME_OR_PATH];
+    assert(nameOrPath === undefined || typeof nameOrPath === "string");
```
coder11, line 68: I was wondering if it's possible to force parsed args (`args` variable) to have the right type?

coder11 (approved):

PR #30, opened 2022-06-19
remove brackets, update version, update npm cmd

boneyard93501:
```
fluence (--version)
zsh: unknown file attribute: v
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,11 +30,11 @@ Don't name arguments or flags with names that contain underscore symbols, becaus
 
 <!-- usage -->
 ```sh-session
-$ npm install -g @fluencelabs/cli
+$ npm install -location=global @fluencelabs/cli
 $ fluence COMMAND
 running command...
-$ fluence (--version)
-@fluencelabs/cli/0.0.2 linux-x64 node-v16.14.0
+$ fluence --version
+@fluencelabs/cli/0.1.3 darwin-x64 node-v16.15.1
 $ fluence --help [COMMAND]
 USAGE
   $ fluence COMMAND


README.md
```diff
@@ -30,11 +30,11 @@ Don't name arguments or flags with names that contain underscore symbols, becaus
 
 <!-- usage -->
 ```sh-session
-$ npm install -g @fluencelabs/cli
+$ npm install -location=global @fluencelabs/cli
```
shamsartem, line 33: all this text is auto-generated so it won't be good to change it like that. But I still wanna know why you wanted to change this line. npm i -g something seems so to be the basic way of installing global npm dependencies. Doesn't it work for you? I asked guys with mac-s to check and I think it worked for them

README.md
```diff
@@ -30,11 +30,11 @@ Don't name arguments or flags with names that contain underscore symbols, becaus
...
 $ fluence COMMAND
 running command...
-$ fluence (--version)
-@fluencelabs/cli/0.0.2 linux-x64 node-v16.14.0
+$ fluence --version
```
shamsartem, line 36: I think creators of the framework decided this auto-generated makes sense because you can also just say `fluence` (without using --version flag) and it will show you the help text, so in my personal opinion it's ok to leave it as it is

README.md
```diff
@@ -30,11 +30,11 @@ Don't name arguments or flags with names that contain underscore symbols, becaus
...
+@fluencelabs/cli/0.1.3 darwin-x64 node-v16.15.1
```
shamsartem, line 37: this one is auto-generated as well as the others (you can spot html comments  and simmilar - everything between these tags is auto-generated. We can remove the tags if we want to but we need to decide on that)

so this exact line is just an example. It depends on the machine that you use. I see 
@fluencelabs/cli/0.1.2 linux-x64 node-v16.15.0

and the version number should be changed in package.json - that's where we should change it instead, I totally forgot to keep doing that

shamsartem (changes requested):
If you think this part of the README is really bad or misleading we can remove usage and usagestop html comments and write this info ourselves but I think we need to just change version in package.json and that's it, everything else is pretty ok in my opinion as it is

README.md
boneyard93501: auto-generation is all good if it's correct. clearly
```bash
fluence (--version)
zsh: unknown file attribute: v```
```
is not correct and shouldn't be propagated.

```bash
npm -g install @fluencelabs/aqua
npm WARN config global `--global`, `--local` are deprecated. Use `--location=global` instead.
```
no point propagating deprecated ways

shamsartem: cool. didn't see this warning ever before. If you don't wanna wait until they change it we should remove these tags and write this ourselves

shamsartem:
I will just close this PR for now. I will update version inside package.json myself and if you think (--version) stuff is too confusing let's open a separate PR and remove tags for auto-generation and replace with a different text

PR #34, opened 2022-08-01
Fluence CLI latest changes

shamsartem:


src/lib/helpers/getHashOfString.ts
```diff
@@ -13,5 +13,3 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
```
coder11, line 15: MB just remove the file?

src/commands/deploy.ts
```diff
@@ -235,32 +578,46 @@ const deployService = async ({
  * @param param0 Readonly<{ deployServiceOptions: DeployServiceOptions; count: number; commandObj: CommandObj;}>
  * @returns Promise<DeployedServiceConfig[] | null>
  */
+/*
 const deployServices = async ({
```
coder11, line 582: Commented out code: either remove or add a note explaining why does it have to be left this way (and when it will be fixed)

src/commands/module/add.ts
```diff
@@ -0,0 +1,146 @@
...
+
+const PATH_OR_URL = "PATH | URL";
+const NAME_FLAG_NAME = "name";
+
+export default class Add extends Command {
+  static override description = `Add module to ${color.yellow(
+    SERVICE_CONFIG_FILE_NAME
+  )}`;
+  static override examples = ["<%= config.bin %> <%= command.id %>"];
+  static override flags = {
```
coder11, line 45: Linter prints an error about unsafe assingment of `any` value

src/commands/module/add.ts
```diff
@@ -0,0 +1,146 @@
...
+    const { args, flags } = await this.parse(Add);
+    const isInteractive = getIsInteractive(flags);
+    const pathToModule: unknown =
+      args[PATH_OR_URL] ??
+      (await input({
+        isInteractive,
+        message: "Enter path to a module or url to .tar.gz archive",
+      }));
+    assert(typeof pathToModule === "string");
+    const serviceNameOrPath =
```
coder11, line 74: Here as well

src/commands/module/add.ts
```diff
@@ -0,0 +1,146 @@
...
+      const serviceGet = fluenceConfig?.services[serviceNameOrPath]?.get;
+      assert(typeof serviceGet === "string");
+      servicePath = serviceGet;
+    }
+    if (isUrl(servicePath)) {
+      this.error(
+        `Can't modify downloaded service ${color.yellow(servicePath)}`
+      );
+    }
+    const serviceConfig = await initServiceConfig(
```
coder11, line 94: And here

src/commands/module/add.ts
```diff
@@ -0,0 +1,146 @@
...
+      this
+    );
+    if (serviceConfig === null) {
+      this.error(
+        `Directory ${color.yellow(servicePath)} does not contain ${color.yellow(
+          SERVICE_CONFIG_FILE_NAME
+        )}`
+      );
+    }
+    const moduleName =
```
coder11, line 105: Here

src/commands/module/add.ts
```diff
@@ -0,0 +1,146 @@
...
+        `You already have ${color.yellow(moduleName)} in ${color.yellow(
+          SERVICE_CONFIG_FILE_NAME
+        )}. Provide a unique name for the new module using ${color.yellow(
+          `--${NAME_FLAG_NAME}`
+        )} flag or edit the existing module in ${color.yellow(
+          SERVICE_CONFIG_FILE_NAME
+        )} manually`
+      );
+    }
+    serviceConfig.modules = {
```
coder11, line 128: And here

src/lib/configs/project/service.ts
```diff
@@ -0,0 +1,161 @@
...
+  getConfigDirPath: (): string => configDirPath,
+  examples,
+});
+
+export const initServiceConfig = (
+  configDirPath: string,
+  commandObj: CommandObj
+): Promise<InitializedConfig<LatestConfig> | null> =>
+  getConfigInitFunction(getInitConfigOptions(configDirPath))(commandObj);
+export const initReadonlyServiceConfig = (
```
coder11, line 134: new line

src/lib/configs/project/service.ts
```diff
@@ -0,0 +1,161 @@
...
+  configDirPath: string,
+  commandObj: CommandObj
+): Promise<InitializedReadonlyConfig<LatestConfig> | null> =>
+  getReadonlyConfigInitFunction(getInitConfigOptions(configDirPath))(
+    commandObj
+  );
+const getDefault: (
```
coder11, line 141: And new line here

src/lib/configs/project/service.ts
```diff
@@ -0,0 +1,161 @@
...
+  (relativePathToFacade: string): GetDefaultConfig<LatestConfig> =>
+  (): LatestConfig => ({
+    version: 0,
+    modules: {
+      [FACADE_MODULE_NAME]: {
+        get: relativePathToFacade,
+      },
+    },
+  });
+export const initNewReadonlyServiceConfig = (
```
coder11, line 153: And one more new line. Is it possible to add a linter rule to force separating top-level declarations?

src/lib/rust.ts
```diff
@@ -0,0 +1,261 @@
...
+  commandObj: CommandObj;
+  message: string;
+}): Promise<string> =>
+  execPromise(
+    `${CARGO}${
+      isNightly === true ? " +nightly" : ""
+    } install ${packageName} ${unparseFlags(
+      {
+        version,
+        ...(isGlobalDependency === true
```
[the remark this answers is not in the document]
shamsartem: Because I think type coercion in JS is a great source of bugs. So I disallow it in a way with linter rules. isGlobalDependency parameter can be undefined, and we know undefined is falsy, but depending on the context we might want to do different stuff depending on whether isGlobalDependency is undefined or false. In this case we don't care, so we explicitly write that we care only about `true` case

src/lib/multiaddr.ts
```diff
@@ -16,25 +16,63 @@
...
+  const randomRelayAddr = addrs[randomIndex];
   assert(randomRelayAddr !== undefined);
   return randomRelayAddr;
 };
 
-export const getRandomRelayId = (): string => {
-  const largestIndex = defaultRelayIds.length - 1;
+const getIds = (nodes: Array<string>): Array<string> =>
+  nodes.map((addr): string => {
+    const addrParts = addr.split("/");
```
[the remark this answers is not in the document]
shamsartem: I've seen it in fluencejs and wanted to add to here as well, but totally forgot that. Thanks for catching that!

src/lib/execPromise.ts
```diff
@@ -45,26 +40,26 @@ export const execPromise = (
...
+            )}\nIt's best to just try again or increase timeout using ${color.yellow(
+              `--${TIMEOUT_FLAG_NAME}`
+            )} flag\n${failedCommandText}`
+          )
+        );
+      }, timeout);
 
     const childProcess = exec(command, (error, stdout, stderr): void => {
-      clearTimeout(execTimeout);
+      execTimeout !== false && clearTimeout(execTimeout);
```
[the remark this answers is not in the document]
shamsartem: First of all I will change that to an explicit if statement. Maybe we should have a linter rule that disallows this && operator stuff

Here is an exact situation as I mentioned in a previous comment. execTimeout variable can be either NodeJS.Timeout or false. In nodejs it's fine but in browser timeout can be a number, so we could theoretically shoot ourselves in the foot if timeout is equal to 0. Safer to be explicit, even though it requires us to write a bit more

src/lib/helpers/getHashOfString.ts
shamsartem: Missed it. Thanks!

src/commands/deploy.ts
shamsartem: Also forgot to remove it. Thanks!
I want to write a better deploy by blueprint id in the future by first calculating it locally and then checking if there is already such a blueprint id on the peer and only then deploying

src/lib/configs/project/service.ts
shamsartem: what's the problem?

shamsartem: You want to add new line before export, so it looks nicer?

coder11 (approved):

PR #38, opened 2022-08-09
remove cargo-generate

shamsartem:


src/commands/module/new.ts
```diff
@@ -62,14 +61,51 @@ export const generateNewModule = async (
...
+name = "${name}"
+version = "0.1.0"
+edition = "2018"
+
+[[bin]]
+name = "${name}"
+path = "src/main.rs"
+
+[dependencies]
+marine-rs-sdk = "0.6.15"
```
coder11, line 107: I think we should extract these versions to `const.ts` file

coder11 (approved):

src/commands/module/new.ts
shamsartem: Yeah, I should totally do it. Thanks

PR #41, opened 2022-08-10
Add ability to selectively remove and deploy services

shamsartem:


src/commands/deploy.ts
```diff
@@ -123,54 +127,59 @@ export default class Deploy extends Command {
...
-      await removeApp({
+    let appConfig = await initAppConfig(this);
+
+    if (
+      appConfig !== null &&
+      (isInteractive
+        ? await confirm({
+            isInteractive,
+            message:
+              "Do you want to remove some of the previously deployed services?",
```
coder11, line 138: Can we specify the exact services that are going to be removed?

shamsartem: If you answer yes, then it will ask about each of the services interactively. A bit annoying but should at least temporarily solve the issue

coder11: Let's specify that the user will be prompted about each service in the message.

coder11 (approved):

PR #42, opened 2022-08-12
Add deploy strategies

shamsartem:


src/lib/multiaddr.ts
```diff
@@ -77,3 +81,34 @@ export const getRandomRelayId = (relays: Relays): string => {
...
+  count = 1
+): Array<string> => {
+  const addrs = resolveAddrs(relays);
+  const ids = getIds(addrs);
+  return getEvenlyDistributedIdsFromTheList(ids, count);
+};
+
+const offsets = new Map<string, number>();
+
+export const getEvenlyDistributedIdsFromTheList = (
```
coder11, line 96: I think it could be written simpler if you make a function which generates a list of randomly distributed indices. Something like:

```
const evenlySpreadIndices = (count, total) => 
[...Array(count).keys()]
   .map(x => x + 1)
   .map(x => total / x))
   .map(x => x - 1)
   .map(Math.round);
```

And then use these indices to pick corresponding items from the collection

src/lib/configs/project/fluence.ts
```diff
@@ -62,12 +63,19 @@ const configSchemaV0: JSONSchemaType<ConfigV0> = {
   required: ["version", "services"],
 };
 
+export const DISTRIBUTION_EVEN = "even";
+export const DISTRIBUTION_RANDOM = "random";
+export const DISTRIBUTIONS = [DISTRIBUTION_EVEN, DISTRIBUTION_RANDOM] as const;
```
coder11, line 68: Why not use `enum` for this?

src/commands/deploy.ts
```diff
@@ -169,42 +179,39 @@ export default class Deploy extends Command {
...
-          secretKey: keyPair.secretKey,
-          aquaCli,
-          timeout: flags[TIMEOUT_FLAG_NAME],
-          tmpDeployJSONPath,
-          commandObj: this,
-          doDeployAll,
-          isInteractive,
-        });
+      // eslint-disable-next-line no-await-in-loop
+      const res = await deployService({
```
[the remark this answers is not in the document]
shamsartem: Because it's unstable this way for some reason, often fails. Something to look into in the future

src/lib/multiaddr.ts
shamsartem: Not completely sure what you are suggesting, exactly.
In my implementation, I also have this offset. This way things can be distributed real evenly across deploys,
e.g. without that, each deploy will have the same peer ID as the first peer ID

src/lib/configs/project/fluence.ts
shamsartem: There are multiple reasons why unions are superior to enums. Basically, the only real feature of enums (some people think it's one of their main flaws) is that they are opaque (means that if you have two of exactly the same enum structurally but defined in different modules - they will not be the same)

Please take a look at these two articles which completely persuaded me to stop using enums:
https://stackoverflow.com/questions/40275832/typescript-has-unions-so-are-enums-redundant
https://www.bam.tech/article/should-you-use-enums-or-union-types-in-typescript

src/commands/deploy.ts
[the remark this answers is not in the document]
coder11: Let's add a comment about that

src/lib/configs/project/fluence.ts
coder11: Sounds reasonable. Although I would put more points for rename refactoring part :)

src/commands/deploy.ts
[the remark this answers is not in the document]
shamsartem: Done. Also, when user requests, we interactively ask about each deploy. Added comment about that as well

src/lib/configs/project/fluence.ts
shamsartem: As far as I can see now, things changed and unions are pretty refactorable as well

src/lib/multiaddr.ts
coder11: Could you please add more comments about the algorithm, namely

1. Describe what is the algorithm and why do we need it
2. Add a note that the function is not pure, and what side effects are

shamsartem: Done

coder11 (approved):
Great job!

PR #43, opened 2022-08-30
Small fixes

shamsartem:
Prompt user for module path when running fluence module new

Provide default secret key automatically when running fluence run

Generate .vscode/settings.json with relative import path and remove it from .gitignore because with VSCode aqua extension supporting relative paths now - it's ok to version control this file


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "aquaSettings.imports": [
+    ".fluence/aqua"
+  ]
+}
```
coder11, line 5: Add newline please

coder11 (approved):

PR #44, opened 2022-09-01
Add key-management and support for key-pairs for deploy, update dependencies, fix paths bug

shamsartem:


src/commands/run.ts
```diff
@@ -72,6 +72,16 @@ export default class Run extends Command {
       helpValue: "<path>",
       multiple: true,
     }),
+    plugin: Flags.string({
+      description: "[experimental] Path to a directory with JS plugins",
```
coder11, line 76: I think we need to a add a link to documentation. This one I suppose: https://doc.fluence.dev/aqua-book/aqua-cli/plugins

coder11 (approved):

src/commands/run.ts
shamsartem: Wouldn't hurt

PR #49, opened 2022-09-21
feat: per-project dependency management

shamsartem:


src/commands/dependency/cargo/install.ts
```diff
@@ -0,0 +1,125 @@
...
+    const { args, flags } = await this.parse(Install);
+    const isInteractive = getIsInteractive(flags);
+    await ensureFluenceProject(this, isInteractive);
+
+    const packageNameAndVersion: unknown =
+      args[PACKAGE_NAME_AND_VERSION_ARG_NAME];
+
+    const fluenceConfig = await initFluenceConfig(this);
+
+    assert(
```
coder11, line 71: could you add a comment explaining why we are doing this check

src/commands/dependency/cargo/install.ts
```diff
@@ -0,0 +1,125 @@
...
+      (packageNameAndVersion === undefined ||
+        typeof packageNameAndVersion === "string") &&
+        fluenceConfig !== null
+    );
+
+    if (packageNameAndVersion === undefined) {
```
coder11, line 77: and this one

src/commands/dependency/npm/install.ts
```diff
@@ -0,0 +1,132 @@
...
+    const { args, flags } = await this.parse(Install);
+    const isInteractive = getIsInteractive(flags);
+    await ensureFluenceProject(this, isInteractive);
+
+    const packageNameAndVersion: unknown =
+      args[PACKAGE_NAME_AND_VERSION_ARG_NAME];
+
+    const fluenceConfig = await initFluenceConfig(this);
+
+    assert(
```
coder11, line 70: and these two

coder11 (approved):

src/lib/rust.ts
```diff
@@ -118,124 +113,129 @@ const isRustInstalled = async (): Promise<boolean> => {
...
-  isNightlyX86?: true;
+  toolchain?: string;
 };
 
-export const cargoDependencies: Record<CargoDependency, CargoDependencyInfo> = {
+export const fluenceCargoDependencies: Record<string, CargoDependencyInfo> = {
   [MARINE_CARGO_DEPENDENCY]: {
     recommendedVersion: MARINE_RECOMMENDED_VERSION,
-    packageName: MARINE_CARGO_DEPENDENCY,
-    isNightlyX86: true,
```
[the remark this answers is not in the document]
shamsartem: Not removing but replacing it with a toolchain flag, which is more flexible. People would be able to install other dependencies with different toolchaing if they want to

PR #59, opened 2022-10-28
feat: add templates

shamsartem:
fix: fluence run --dry

### DIFF
--- a/src/lib/const.ts
+++ b/src/lib/const.ts
@@ -119,10 +119,7 @@ export const TIMEOUT_FLAG = {
   }),
 } as const;
 
-export const TEMPLATE_MINIMAL = "minimal";
-export const TEMPLATE_TS = "ts";
-export const TEMPLATE_JS = "js";
-export const templates = [TEMPLATE_MINIMAL, TEMPLATE_TS, TEMPLATE_JS] as const;
+export const templates = ["minimal", "ts", "js"] as const;
 export type Template = typeof templates[number];
 export const isTemplate = (template: unknown): template is Template =>
   // eslint-disable-next-line unicorn/prefer-includes
--- a/src/lib/const.ts
+++ b/src/lib/const.ts
@@ -135,13 +135,30 @@ export const AQUA_LIB_NPM_DEPENDENCY = "@fluencelabs/aqua-lib";
 
 export const MAIN_AQUA_FILE_CONTENT = `module Main
 
--- import App from "deployed.app.aqua"
 import "@fluencelabs/aqua-lib/builtin.aqua"
 
--- export App
+
+-- Uncomment the following when you deploy your app with Adder service:
+
+-- import App from "deployed.app.aqua"
+-- import Adder from "services/adder.aqua"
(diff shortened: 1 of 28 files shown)


src/lib/paths.ts
```diff
@@ -120,17 +122,51 @@ export const ensureFluenceAquaServicesDir = async (): Promise<string> =>
 export const ensureFluenceAquaDeployedAppPath = async (): Promise<string> =>
   path.join(await ensureFluenceAquaDir(), DEPLOYED_APP_AQUA_FILE_NAME);
 
-export const ensureFluenceJSAppPath = (fluenceJSDir: string): string =>
-  path.join(fluenceJSDir, APP_JS_FILE_NAME);
+// JS
 
-export const ensureFluenceJSDeployedAppPath = (fluenceJSDir: string): string =>
-  path.join(fluenceJSDir, DEPLOYED_APP_JS_FILE_NAME);
+export const ensureFluenceJSAppPath = async (
```
coder11, line 127: Are these paths configurable? I mean JS and TS destination paths.

src/lib/const.ts
```diff
@@ -142,52 +160,52 @@ import "@fluencelabs/aqua-lib/builtin.aqua"
 
 -- import App from "deployed.app.aqua"
 -- import Adder from "services/adder.aqua"
--- export App, add_one
+-- export App, addOne
```
coder11, line 163: If you are doing snake_case to camelCase renaming only for the sake of making generated TS files look nice than I think we are doing it in the wrong place. This should be done inside aqua compiler.

src/lib/paths.ts
coder11: @DieMyst WDYT?

@folex  WDYT about recommended naming notation?

src/lib/const.ts
coder11: @DieMyst WDYT?

@folex WDYT about recommended naming notation?

folex: I like snake case. But I don't care much.

src/lib/paths.ts
shamsartem: now yes

src/lib/const.ts
shamsartem: I think compiler shouldn't do that. It's just camelCase is a lot more popular among js developers so I think people who compile to js would prefer using camelCase themselves

src/lib/paths.ts
coder11: awesome

src/lib/const.ts
coder11: Our built-ins library is using snake case in function naming. So if a developer would be using camelCase they would inevitably face style inconsistencies

coder11 (approved):
Feel free to merge the changes if you need them ASAP. But I would suggest to come to some sort of an agreement on naming style and fix the aqua template accordingly.

shamsartem:
I will merge this for now. If we decide something different - I will just do another PR with a fix

PR #75, opened 2022-12-20
feat: refactor and solve interrupted dependency installation

shamsartem:


src/lib/const.ts
```diff
@@ -182,7 +182,7 @@ export const MAIN_AQUA_FILE_ADD_ONE_COMMENT = aquaComment(
   MAIN_AQUA_FILE_ADD_ONE
 );
 
-export const MAIN_AQUA_FILE_CONTENT = `module Main
+export const MAIN_AQUA_FILE_CONTENT = `aqua Main
```
coder11, line 185: Was it replaced by accident?

coder11 (changes requested):

coder11 (approved):

PR #76, opened 2022-12-27
feat: migrate aqua run to fluence run [DXJ-170]

shamsartem:


src/lib/aqua.ts
```diff
@@ -0,0 +1,87 @@
...
+  AquaConfig,
+  CompilationResult,
+  Aqua,
+  Call,
+  Input,
+  Path,
+} from "@fluencelabs/aqua-api/aqua-api";
+import type { callFunctionImpl } from "@fluencelabs/fluence/dist/internal/compilerSupport/v3impl/callFunction";
+
+export type Data = Parameters<typeof callFunctionImpl>[4];
```
coder11, line 29: We can make a type and export it in FJS for the `data` argument

src/lib/aqua.ts
```diff
@@ -0,0 +1,87 @@
...
+      new Call(funcCall, data, new Input(filePath)),
+      imports,
+      config
+    );
+
+    return result;
+  }
+
+  if (typeof code === "string") {
+    return Aqua.compile(new Input(code), imports, config);
```
coder11, line 82: Does it just call the aqua through the command line?

coder11: Ahh, I see. That's the usage of aqua-api

src/lib/aqua.ts
```diff
@@ -0,0 +1,87 @@
...
+  noXor = false,
+}: {
+  code?: string;
+  filePath?: string;
+  funcCall?: string;
+  data?: Data | undefined;
+} & CommonArgs): Promise<CompilationResult> {
+  const config = new AquaConfig(logLevel, constants, noXor, noRelay);
+
+  if (typeof funcCall === "string" && filePath !== undefined) {
```
coder11, line 71: Could you please elaborate in comments why these checks are made and what user scenarios they match to?

coder11 (approved):

src/lib/aqua.ts
shamsartem: This can be done. Maybe together with changing `callFunctionImpl` to accept an object with some of the fields that are optional, instead of 5 arguments each of which is required

src/lib/aqua.ts
shamsartem: will do

PR #77, opened 2022-12-30
feat(countly): add Countly [DXJ-191]

shamsartem:


README.md
```diff
@@ -91,14 +91,16 @@ Don't name arguments or flags with names that contain underscore symbols, becaus
...
+
 Pull request and release process:
 1. Run `npm run check` to make sure everything ok with the code
-2. Only after that commit your changes to trigger pre-commit hook that updates `README.md`. Read `README.md` to make sure it is correctly updated
-3. Push your changes
-4. Create pull request and merge your changes to `main`
-5. Switch to `main` locally and pull merged changes
-6. Run `git tag -a v0.0.0 -m ""` with version number that you want instead of `0.0.0`
-7. Run `git push origin v0.0.0` with version number that you want instead of `0.0.0` to trigger release
+1. Only after that commit your changes to trigger pre-commit hook that updates `README.md`. Read `README.md` to make sure it is correctly updated
```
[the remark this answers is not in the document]
shamsartem: I changed that so it's more maintainable. It is still rendered correctly, markdown supports this style. Now it's easier to add new intermediate steps

src/countlyInterceptor.js
```diff
@@ -0,0 +1,73 @@
+const {
```
coder11, line 1: Why can't we write this logic in a typescript file?

shamsartem: because we have to import it into a js files without compilation step (into ./bin/dev and ./bin/run). That's how this oclif framework is set up. But this file is still type-checked though

src/lib/countly.ts
```diff
@@ -0,0 +1,102 @@
...
+    }),
+    {}
+  );
+
+/**
+ * Add log that will be sent to Countly together with crash report
+ * @param message - message to be logged
+ * @returns void
+ */
+export const addLog = (message: string): void => {
```
coder11, line 87: I'd suggest to rename to `addCountlyLog` to avoid possible confusion

shamsartem: ok. agree

src/lib/countly.ts
```diff
@@ -0,0 +1,102 @@
...
+    {}
+  );
+
+/**
+ * Add log that will be sent to Countly together with crash report
+ * @param message - message to be logged
+ * @returns void
+ */
+export const addCountlyLog = (message: string): void => {
+  if (Countly.device_id === undefined) {
```
coder11, line 88: Why can't we proceed in that case? What `device_id` mean?

shamsartem: I will add a function for that so it's more clear. The idea is that if there is no device_id - Countly was not initialized

src/countlyInterceptor.js
coder11: But the rest of the project is written in Typescript. It even executes everything through the TS pipeleine (`dev` file):

```
require("ts-node").register({ project });
```

Can't we run this file using the same mechanism?

coder11 (approved):

PR #87, opened 2023-01-11
feat(dependencies): remove required dependencies [DXJ-227]

shamsartem:


src/lib/helpers/generateKeyPair.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+import { KeyPair } from "@fluencelabs/fluence";
+
+import type { ConfigKeyPair } from "../configs/keyPair";
+
+export const generateKeyPair = async (name: string): Promise<ConfigKeyPair> => {
+  const keyPair = await KeyPair.randomEd25519();
+  return {
+    peerId: keyPair.Libp2pPeerId.toB58String(),
```
coder11, line 24: This functionality should really belong to the `KeyPair` package. I'm not sure how and well we should do it, but we should :)

coder11 (approved):

PR #113, opened 2023-01-18
feat(run): use --input flag to search for project root dir path [fixes DXJ-232]

shamsartem:


folex (approved):
Some comments would be nice, otherwise lgtm

PR #123, opened 2023-01-24
chore(ci): Notify on release failures and other small CI changes [fixes FLU-222, FLU-194, FLU-41 and FLU-196]

nahsi:
image


shamsartem (approved):
Thanks

PR #132, opened 2023-02-01
feat: generate app config for new deploy [fixes DXJ-245]

shamsartem:


shamsartem:
A different approach with a separate config was chosen

PR #133, opened 2023-02-03
feat: migrate to es modules [fixes DXJ-252]

shamsartem:


src/lib/prompt.ts
```diff
@@ -16,12 +16,15 @@
 
 import assert from "node:assert";
 
-import color from "@oclif/color";
+import oclifColor from "@oclif/color";
+const color = oclifColor.default;
 import type { JSONSchemaType, ValidateFunction } from "ajv";
-import inquirer, { Answers, DistinctQuestion, Separator } from "inquirer";
+import inquirer, { Answers, DistinctQuestion } from "inquirer";
```
coder11, line 22: Is it possible to write the following instead?
```
import type { Separator } from "inquirer";
```

src/lib/multiaddr.ts
```diff
@@ -22,11 +22,14 @@ import {
   testNet,
   Node,
 } from "@fluencelabs/fluence-network-environment";
-import shuffle from "lodash/shuffle";
+import lodash from "lodash";
```
coder11, line 25: I believe there is a special package for lodash with ESM, https://www.npmjs.com/package/lodash-es
Have you tried using it instead of `lodash`?

src/commands/aqua.ts
```diff
@@ -14,23 +14,29 @@
  * limitations under the License.
  */
 
+import { readdir, readFile, writeFile } from "node:fs/promises";
```
coder11, line 17: I suggest to have a single way to do imports from `fs`
Currently you have two different "ways" to do it:

```
import { readdir, readFile, writeFile } from "node:fs/promises";
```

and

```
import fsPromises from "node:fs/promises";
```

src/lib/init.ts
```diff
@@ -331,12 +330,14 @@ export const initTSorJSProject = async ({
   } else {
     const TS_CONFIG = {
       compilerOptions: {
-        target: "es2016",
-        module: "commonjs",
-        esModuleInterop: true,
-        forceConsistentCasingInFileNames: true,
+        target: "es2022",
```
coder11, line 333: Are you sure we should change to content of the JS template right now before we released the new version of JS Client?

src/commands/aqua.ts
```diff
@@ -190,9 +209,48 @@ export default class Aqua extends Command {
...
+  const outputPath = maybeOutputPath;
+
+  if (path.isAbsolute(outputPath)) {
+    return outputPath;
+  }
+
+  return path.resolve(projectRootDir, outputPath);
+};
+
+const addFileExtensionsInTsFiles = async (outputDirPath: string) => {
```
coder11, line 236: This must be updated in `Aqua` directly instead of trying to overwrite the content of generated files.
Moreover are you sure it's time to migrate JS\TS projects (not the CLI itself, but the generated code) to ESM. We haven't released the new JS Client yet

coder11 (changes requested):

src/lib/multiaddr.ts
shamsartem: no I did not. I will. Thanks

src/commands/aqua.ts
shamsartem: yes I am migrating towards the first one gradually where I find it. The first one is better because it is more concise and imports automatically

src/lib/init.ts
shamsartem: It works like this. Tests make sure it works too

src/commands/aqua.ts
shamsartem: This is a temporary measure so it works at least. I agree with you. Also I think maybe compilation to js should be done by fluence cli and aqua should only be concerned with producing air

coder11 (approved):
Bring the ESM in!

PR #141, opened 2023-02-14
feat: add configs for workers (worker.yaml), direct hosting (hosts.yaml) and deployment result (deployed.yaml). Add command for worker deployment. Add command for showing versions of used dependencies. Fix issue with circular imports [fixes DXJ-246 DXJ-247 DXJ-259 DXJ-249 DXJ-270]

shamsartem:


coder11 (approved):
I would suggest find a way not to commit auto-generated aqua code.

PR #161, opened 2023-02-20
fix: don't print result of aqua call twice, check that workers have services and that peerIds is not an empty array when deploying workers, prettify error messages  [fixes DXJ-301]

shamsartem:


src/commands/deal/deploy.ts
```diff
@@ -130,6 +131,25 @@ export default class Deploy extends BaseCommand<typeof Deploy> {
       workersConfig,
     });
 
+    const errorMessages = uploadArg.workers.reduce<Array<string>>(
+      (acc, { config: { services }, name }) => {
+        if (services.length === 0) {
+          acc.push(
```
coder11, line 137: I don't like the idea of doing statefull operations in reduce. I would generally prefer to do it like this: `[...acc, newItem]` even though it worse performance-wise. I we really need to push performance to it's limits then I would prefer rewriting in into stateful for loop with explicit mutations.

Also, seems like this code could be rewritten into a chain of `filter` and then `map`. Am I missing something?

src/commands/workers/deploy.ts
```diff
@@ -113,6 +116,33 @@ export default class Deploy extends BaseCommand<typeof Deploy> {
       workersConfig,
     });
 
+    const errorMessages = uploadDeployArg.workers.reduce<Array<string>>(
+      (acc, { config: { services }, hosts, name }) => {
+        if (services.length === 0) {
```
coder11, line 121: I think this once can be rewritten as a chain of `map` and `filter` as well. Just return `null` if you don't want an item to be included into resulting collection and then filter out `null` elements

src/commands/workers/upload.ts
```diff
@@ -113,6 +116,33 @@ export default class UPLOAD extends BaseCommand<typeof UPLOAD> {
       workersConfig,
     });
 
+    const errorMessages = uploadArg.workers.reduce<Array<string>>(
+      (acc, { config: { services }, hosts, name }) => {
+        if (services.length === 0) {
```
coder11, line 121: ditto

src/countlyInterceptor.js
```diff
@@ -26,7 +28,17 @@ const handleError = (/** @type {unknown} */ error) => {
       ? error
       : new CLIError(`Error: ${String(error ?? "no error?")}`);
 
-  console.error(err);
+  if ("stack" in err) {
```
coder11, line 31: I suggest to extract this block of code into a helper method. Would be easier to read this way.

coder11 (approved):

src/commands/deal/deploy.ts
shamsartem: I did all of these improvements and we will merge it as a part of the next PR. Thanks

PR #190, opened 2023-02-28
feat(blockchain): Migrate to walletconnect v2

elshan-eth:


src/lib/provider.ts
```diff
@@ -130,14 +106,58 @@ export const getSigner = async (
...
+    assert(typeof bridge === "string");
+    const key = connectionStringUrl.searchParams.get(KEY_QUERY_PARAM_NAME);
+    assert(typeof key === "string");
+    const url = new URL(CLI_CONNECTOR_URL);
+    url.searchParams.set(WC_QUERY_PARAM_NAME, wc);
+    url.searchParams.set(RELAY_QUERY_PARAM_NAME, bridge);
+    url.searchParams.set(KEY_QUERY_PARAM_NAME, key);
+
+    commandObj.log(
+      `To approve transactions with your to your wallet using metamask, open the following url:\n\n${url.toString()}\n\nor go to ${CLI_CONNECTOR_URL} and enter the following connection string there:\n\n${uri}\n`
```
shamsartem, line 129: `with` or `to`?

elshan-eth: I think to

package.json
```diff
@@ -70,6 +70,7 @@
     "node-fetch": "^2.6.9",
     "platform": "^1.3.6",
     "replace-homedir": "^2.0.0",
+    "safe-json-utils": "^1.1.1",
```
shamsartem, line 73: I don't see you ever using this or lokijs dependency. Are you sure they both are necessary? If all or some are not please remove and check it's still working

elshan-eth: I removed safe-json-utils. 

lokijs needs for WalletConnect

elshan-eth: https://docs.walletconnect.com/2.0/javascript/guides/nodejs

package.json
```diff
@@ -106,6 +105,7 @@
     "husky": "^8.0.3",
     "jest": "^29.4.1",
     "lint-staged": "^13.1.0",
+    "lokijs": "^1.5.12",
```
shamsartem, line 108: If you are sure lokijs is required when user runs cli on his computer - please remove this dependency and and install it again without -D flag, because currently it's in "devDependencies" section in package.json which means it should go to the package itself and is only used during the development process

elshan-eth: To use WalletConnect server side, you'll need to install the "lokijs" dependency. If you are seeing this error during a build / in an SSR environment, you can add "lokijs" as a devDependency to make this error go away.

elshan-eth: image
It's in wallet connect codes

elshan-eth: Is it better to move it from devDeps to Deps?

shamsartem (approved):
Thanks

PR #202, opened 2023-03-07
feat: minor improvements and refactoring [fixes DXJ-331]

shamsartem:


coder11 (approved):
Please consider migrating from "old fluence js" ASAP since we now have a working Node.js implementation of JS Client

PR #204, opened 2023-03-10
chore(e2e): Use fluencelabs/rust-peer image

nahsi:


shamsartem (changes requested):
We should solve this thing that docs are generated with the commit. Because you are not the first one who accidentally wiped the docs out because I suppose you didn't do `npm i` before committing so it couldn't build so it couldn't generate proper docs. Please rollback docs

shamsartem (approved):

PR #216, opened 2023-03-21
feat: support spells [fixes DXJ-340 DXJ-341]

shamsartem:
* add `fluence spell new` command
* add spell.yaml config with validation
* allow listing spells in fluence.yaml with validation
* turn aqua run by default
* various improvements and fixes


src/lib/configs/project/spell.ts
```diff
@@ -0,0 +1,294 @@
...
+    type: "object",
+    nullable: true,
+    description: `Trigger the spell execution periodically. If you want to disable this property by overriding it in ${FLUENCE_CONFIG_FILE_NAME} - pass empty config for it like this: \`clock: {}\``,
+    properties: {
+      periodSec: {
+        type: "number",
+        description:
+          "How often the spell will be executed. If set to 0, the spell will be executed only once. If this value not provided at all - the spell will never be executed",
+        minimum: 0,
+        maximum: U32_MAX,
```
kmd-fl, line 92: I wonder if it's possible to set the maximum here to `MAX_PERIOD_SEC`.

src/lib/configs/project/spell.ts
```diff
@@ -0,0 +1,294 @@
...
+      "Path to Aqua file which contains an Aqua function that you want to use as a spell",
+  },
+  function: {
+    type: "string",
+    description: "Name of the Aqua function that you want to use as a spell",
+  },
+  initArgs: {
+    type: "object",
+    description:
+      "A map of Aqua function arguments names as keys and arguments values as values. They will be passed to the spell function. Also they will be stored in the key-value storage for this particular spell. It's possible to change argument values inside the spell itself if you return a new map of new arguments from the spell function. Also it's possible to update these values from another spell using Spell.set_string method",
```
kmd-fl, line 79:
```suggestion
      "A map of Aqua function arguments names as keys and arguments values as values. They will be passed to the spell function and will be stored in the key-value storage for this particular spell. It's possible to change argument values inside the spell itself if you return a new map of new arguments from the spell function or using Spell.set_string method",
```

kmd-fl: But maybe explaining how to modify these arguments isn't a good idea here. I think this explanation might be confusing without examples. 

```suggestion
      "A map of Aqua function arguments names as keys and arguments values as values. They will be passed to the spell function and will be stored in the key-value storage for this particular spell.",
```

src/lib/configs/project/spell.ts
```diff
@@ -0,0 +1,294 @@
...
+  },
+  clock: {
+    type: "object",
+    nullable: true,
+    description: `Trigger the spell execution periodically. If you want to disable this property by overriding it in ${FLUENCE_CONFIG_FILE_NAME} - pass empty config for it like this: \`clock: {}\``,
+    properties: {
+      periodSec: {
+        type: "number",
+        description:
+          "How often the spell will be executed. If set to 0, the spell will be executed only once. If this value not provided at all - the spell will never be executed",
```
kmd-fl, line 90: > If this value _is_ not provided at all - the spell will never be executed

Is this part of fcli logic?

Since you cannot _not_ provide period to the rust-peer.

src/lib/configs/project/spell.ts
```diff
@@ -0,0 +1,294 @@
...
+) =>
+  validateBatch(
+    config.clock?.startTimestamp !== undefined &&
+      config.clock?.startDelaySec !== undefined
+      ? `You can't specify both 'startTimestamp' and 'startDelaySec' properties`
+      : true,
+
+    config.clock?.endTimestamp !== undefined &&
+      config.clock?.endDelaySec !== undefined
+      ? `You can't specify both 'endTimestamp' and 'endDelaySec' properties`
```
kmd-fl, line 197: Inconsistent with the doc above. The same with `startTimestamp`.
> This property overrides `endTimestamp` if both are specified",

src/lib/configs/project/spell.ts
```diff
@@ -0,0 +1,294 @@
...
+          return `Start time must be earlier than end time. Got: start time ${startSec}, end time ${endSec}`;
+        }
+
+        const currentSec = getDateSec(new Date());
+
+        if (endSec !== 0 && endSec < currentSec) {
+          return `End time must be later than current time. Got: current time ${currentSec}, end time ${endSec}`;
+        }
+
+        if (endSec - startSec > MAX_PERIOD_SEC) {
```
kmd-fl, line 226: The period to which this restriction applies is `periodSec` field, not `endSec - startSec`.

src/lib/configs/project/spell.ts
```diff
@@ -0,0 +1,294 @@
...
+    type: "string",
+    description: "Name of the Aqua function that you want to use as a spell",
+  },
+  initArgs: {
+    type: "object",
+    description:
+      "A map of Aqua function arguments names as keys and arguments values as values. They will be passed to the spell function. Also they will be stored in the key-value storage for this particular spell. It's possible to change argument values inside the spell itself if you return a new map of new arguments from the spell function. Also it's possible to update these values from another spell using Spell.set_string method",
+    nullable: true,
+  },
+  clock: {
```
kmd-fl, line 82: I didn't find the code that converts this config to actual `TriggerConfig`, so I hope it converts to the same JSON described in the docs and in `spell/spell_service.aqua`.

src/lib/configs/project/spell.ts
shamsartem: Yes it's part of the logic I wrote. The reason is - I don't know any other good way to succinctly express that you wanna override the property in yaml (json form) - this enables passing empty object for clock property in fluence.yaml spell overrides to express that you don't wanna run clock at all (even though spell definition includes it)

src/lib/configs/project/spell.ts
shamsartem: Oh damn thank you for noticing it. I changed it and totally forgot about it. I personally think it would be better to have an error so it's more explicit what is happening

src/lib/configs/project/spell.ts
shamsartem: I think it would be valid to do that. Thank for suggestion

src/lib/configs/project/spell.ts
shamsartem: Oh clearly. Thanks!

src/lib/configs/project/spell.ts
shamsartem: Yes! You can check it out here https://github.com/fluencelabs/fluence-cli/pull/216/files#diff-56fecce60262118f12e6aae7d1351542e57b8ccc7c018cb9b11cfe947cb338e7R294

kmd-fl (approved):

PR #226, opened 2023-04-04
feat: update fluence js-client and remove legacy deploy tests

shamsartem:


src/commands/workers/deploy.ts
```diff
@@ -64,8 +65,8 @@ export default class Deploy extends BaseCommand<typeof Deploy> {
       true
     );
 
-    const fluenceClient = await initFluenceClient(flags, fluenceConfig);
-    doRegisterIpfsClient(fluenceClient, flags["off-aqua-logs"]);
+    await initFluenceClient(flags, fluenceConfig);
+    doRegisterIpfsClient(flags["off-aqua-logs"]);
```
folex, line 69: It's kinda strange that `ipfs client` receives flag `off-aqua-logs`. What is happening here? :)

src/commands/workers/upload.ts
```diff
@@ -58,8 +58,8 @@ export default class UPLOAD extends BaseCommand<typeof UPLOAD> {
       true
     );
 
-    const fluenceClient = await initFluenceClient(flags, fluenceConfig);
-    doRegisterIpfsClient(fluenceClient, flags["off-aqua-logs"]);
+    await initFluenceClient(flags, fluenceConfig);
+    doRegisterIpfsClient(flags["off-aqua-logs"]);
```
folex, line 62: ditto

src/lib/jsClient.ts
```diff
@@ -39,16 +41,17 @@ export const initFluenceClient = async (
     "off-aqua-logs": offAquaLogs,
   }: FluenceClientFlags & KeyPairFlag & OffAquaLogsFlag,
   maybeFluenceConfig: FluenceConfig | null
-): Promise<FluencePeer> => {
-  const client = createClient();
+): Promise<void> => {
   const relay = maybeRelay ?? getRandomRelayAddr(maybeFluenceConfig?.relays);
 
+  commandObj.log(`Fluence js-client connecting to: ${color.yellow(relay)}`);
```
folex, line 47: Maybe like that? Not sure what user should understand from `Fluence js-client`
```suggestion
  commandObj.log(`Connecting to: ${color.yellow(relay)}`);
```

Also, maybe let's mark log-level? Is there something like `commandObj.info`?

src/commands/workers/deploy.ts
folex: Ah, I got it. `IpfsClient` is a service, and there's a `log` function in there. 

So some scripts do `IpfsClient.log`? That's pretty strange still. Maybe no scripts do that, and we can remove it safely?

shamsartem: What do you wanna remove? logs in ipfs client use commandObj.log. Yeah a reused off-aqua-logs for that as well so if this flag is used there would be no such logs printed as a result of aqua execution

src/lib/jsClient.ts
shamsartem: Let's do it like you suggested. I wanna rework logging the same way it's done in js-client so it's consistent. Basically I wanna remove all such flags and show this logs when setting a topic in env variable

src/commands/workers/deploy.ts
folex: I guess just renaming to `off-ipfs-client-logs` would make sense

shamsartem: Discussed in DM to turn them off for now completely. Done

PR #230, opened 2023-04-07
fix(spell): @fluencelabs/installation-spell 0.5.9

folex:
also print error in run.js

### DIFF
--- a/bin/run.js
+++ b/bin/run.js
@@ -23,4 +23,9 @@ import { createErrorPromise } from "../dist/countlyInterceptor.js";
 oclif
   .run(process.argv.slice(2), import.meta.url)
   .then(oclif.flush)
-  .catch(/** @param {unknown} error */ (error) => createErrorPromise(error));
+  .catch(
+    /** @param {unknown} error */ (error) => {
+      console.error(error);
+      return createErrorPromise(error);
+    }
+  );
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -76,7 +76,7 @@
   },
   "devDependencies": {
     "@adobe/jsonschema2md": "^7.1.5",
-    "@fluencelabs/installation-spell": "^0.5.7",
+    "@fluencelabs/installation-spell": "^0.5.9",
```
shamsartem, line 79: did installation spell air change? if yes - it needs to be manually updated (unstil new release of aqua this is still part of the process for now)

bin/run.js
```diff
@@ -23,4 +23,9 @@ import { createErrorPromise } from "../dist/countlyInterceptor.js";
 oclif
   .run(process.argv.slice(2), import.meta.url)
   .then(oclif.flush)
-  .catch(/** @param {unknown} error */ (error) => createErrorPromise(error));
+  .catch(
+    /** @param {unknown} error */ (error) => {
+      console.error(error);
```
shamsartem, line 28: It's weird for me that error doesn't still fucking print. We can't do it like you did it here because this will show stack traces for ALL errors, which we don't want to do. I will try to reproduce this and figure out the problem and solve it correctly

package.json
shamsartem: ok it didn't change, I checked

bin/run.js
shamsartem: fixed this correctly, I think

shamsartem (approved):

PR #234, opened 2023-04-11
feat: update dependencies

shamsartem:


test/helpers.ts
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import "@total-typescript/ts-reset";
```
folex, line 17: I don't see it being used in code 🤔

folex (approved):

test/helpers.ts
shamsartem: it's an override for builtin typescript types

e.g. JSON.parse normally returns `any` but it should return `unknown` - this import makes it return unknown among other things

folex: got it, thx

PR #238, opened 2023-04-12
feat: add -g flag for global dependency management and warn about using dependency overrides [fixes DXJ-362 DXJ-329]

shamsartem:


src/commands/dependency/reset.ts
```diff
@@ -24,17 +34,90 @@ export default class Reset extends BaseCommand<typeof Reset> {
...
+    const { maybeFluenceConfig, flags } = await initCli(
       this,
-      await this.parse(Reset),
-      true
+      await this.parse(Reset)
     );
 
-    delete fluenceConfig.dependencies;
-    await fluenceConfig.$commit();
+    if (flags.global) {
```
folex, line 48: Let's move this into a separate function. So that there are two functions: `resetGlobal` and `resetLocal`.

Single "do-it-all" function is very hard to understand.

src/lib/lifeCycle.ts
```diff
@@ -50,11 +50,12 @@ import {
 } from "./paths.js";
 import { confirm } from "./prompt.js";
 
-const ensureUserConfig = async (): Promise<UserConfig> => {
-  const userConfig = await initUserConfig();
+const ensureUserConfig = async (): Promise<void> => {
+  const maybeUserConfig = await initUserConfig();
```
folex, line 54: What does `initUserConfig` do? Does it _initialize_ config in a sense of creating it? Or does it read from the file?

If read, then why is it called `init`? Very confusing, every time I read it.

src/lib/npm.ts
```diff
@@ -168,6 +161,7 @@ export const ensureNpmDependency = async ({
   });
 
   if (
+    global === false &&
     maybeFluenceConfig !== null &&
     version !==
       (maybeFluenceConfig?.dependencies?.npm?.[name] ??
```
folex, line 167: This condition is something :D

Can you simplify it? Move calculations to separate variables, add comments, etc

src/lib/rust.ts
```diff
@@ -333,9 +326,10 @@ export const ensureCargoDependency = async ({
   }
 
   if (
+    global === false &&
     maybeFluenceConfig !== null &&
     version !==
-      (maybeFluenceConfig?.dependencies?.npm?.[name] ??
+      (maybeFluenceConfig?.dependencies?.cargo?.[name] ??
         (isFluenceCargoDependency(name) ? versions.cargo[name] : undefined))
```
folex, line 333: This must be simplified

src/lib/rust.ts
```diff
@@ -344,6 +338,17 @@ export const ensureCargoDependency = async ({
       packageManager: "cargo",
       version,
     });
+  } else if (
+    global === true &&
+    version !==
+      (userConfig.dependencies?.cargo?.[name] ??
+        (isFluenceCargoDependency(name) ? versions.cargo[name] : undefined))
```
folex, line 345: ditto

src/commands/dependency/reset.ts
shamsartem: I think this is just unecessary and inconcistent. -g or --global flag is used in other commands as well and I don't think it is that hard wrap your head around. Having one more command here will not decrese complexity in any way in my opinion - it will just spread the complexity around a little bit

src/lib/lifeCycle.ts
```diff
@@ -138,15 +138,14 @@ export async function initCli<
     );
   }
 
-  const [userConfig, maybeFluenceConfig] = await Promise.all([
-    ensureUserConfig(),
+  const [maybeFluenceConfig] = await Promise.all([
     initFluenceConfig(),
+    ensureUserConfig(),
   ]);
```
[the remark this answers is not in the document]
shamsartem: I can leave a comment. I use Promise.all to just batch the functions that don't depend on each other. ensureUserConfig is a bit different from other stuff because if this config doesn't exist - we have to ask the user for countly concent which is an async operation as well. User config acts like a singleton so we do not destructure it in array cause it is available everywhere in the program anyway

src/lib/npm.ts
shamsartem: agree. This is completely unredable. Will improve

src/lib/npm.ts
```diff
@@ -179,6 +173,17 @@ export const ensureNpmDependency = async ({
       packageManager: "npm",
       version,
     });
+  } else if (
+    global === true &&
+    version !==
+      (userConfig.dependencies?.npm?.[name] ??
+        (isFluenceNPMDependency(name) ? versions.npm[name] : undefined))
```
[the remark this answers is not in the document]
shamsartem: agree too. Will improve

src/lib/lifeCycle.ts
```diff
@@ -50,11 +50,12 @@ import {
...
 
-const ensureUserConfig = async (): Promise<UserConfig> => {
-  const userConfig = await initUserConfig();
+const ensureUserConfig = async (): Promise<void> => {
+  const maybeUserConfig = await initUserConfig();
 
-  if (userConfig !== null) {
-    return userConfig;
+  if (maybeUserConfig !== null) {
+    setUserConfig(maybeUserConfig);
```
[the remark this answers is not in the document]
shamsartem: It works because before each command execution we do init. This is ok, but not testable. For testability it would be better to figure out a good pattern for dependency injection here, which I will definitly will have to do when I finnaly get to work on testing

src/lib/lifeCycle.ts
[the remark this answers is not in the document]
shamsartem: Added a comment

src/commands/dependency/reset.ts
shamsartem: also all this dependency management I expect to be more for advanced users, especially untill we have proper dependency management on aqua side. So I expect users to mostly use the stuff we provided with the versions we provided. I unfortunately don't expect much publishing of their own libraries for now that they will themselves use

PR #279, opened 2023-06-07
feat: update deployed workers structure in aqua to use optional values [fixes DXJ-382]

shamsartem:


src/lib/build.ts
```diff
@@ -281,9 +283,11 @@ export const build = async ({
     )
   );
 
-  startSpinner("Making sure all services are built");
-  await buildModules([...mapOfModuleConfigs.values()], marineCli);
-  stopSpinner();
+  if (serviceInfos.length > 0) {
+    startSpinner("Making sure all services are built");
+    await buildModules([...mapOfModuleConfigs.values()], marineCli);
+    stopSpinner();
```
folex, line 289: maybe put `stopSpinner` to `finally`? in case `buildModules` throws an error, wouldn't that cause spinner to spin indefinitely?

src/lib/const.ts
```diff
@@ -337,29 +337,29 @@ export const SEPARATOR = `\n\n${color.yellow(
...
+    answers: *Answer
     workers <- resolveSubnetwork(dealId)
     for w <- workers! par:
         on w.metadata.peer_id via w.metadata.relay_id:
             answer <- MyService.greeting("fluence")
-            answers <<- answer
-            print(answer, w.metadata.peer_id)
+            answers <<- Answer(answer=answer, peer=w.metadata.relay_id!)
 
-    Console.print("getting answers...")
```
folex, line 359: Why? I'd keep it

src/lib/deployWorkers.ts
```diff
@@ -622,13 +633,89 @@ const validateWasmExist = async (
   }
 };
 
+const DEAL_TYPE: Deals[string] = {
+  dealId: "",
+  chainNetwork: "testnet",
```
folex, line 638: But that's not TYPE, it's a value 🤔 Why is it called `DEAL_TYPE`?

folex: `Deals[string]` is confusing, too

src/lib/deployWorkers.ts
```diff
@@ -622,13 +633,89 @@ const validateWasmExist = async (
...
     await ensureFluenceAquaWorkersPath(),
     jsToAqua(
-      { deals: workersConfig.deals, hosts: workersConfig.hosts },
+      {
+        deals: makeOptional(
+          Object.keys(workerDealOptionals).length === 0
+            ? null
+            : workerDealOptionals,
+          {
+            [DEFAULT_WORKER_NAME]: makeOptional(null, DEAL_TYPE),
```
folex, line 707: Will it override `DEFAULT_WORKER_NAME` in `workerDealOptionals`?

src/lib/helpers/jsToAqua.ts
```diff
@@ -24,6 +24,43 @@ import { capitalize } from "./capitilize.js";
...
+ * @example
+ * const v = makeOptional(undefined, 1);
+ * // v = { $$optional: 1, $$isNil: true }
+ * @example
+ * const v = makeOptional(null, 1);
+ * // v = { $$optional: 1, $$isNil: true }
+ */
+export const makeOptional = <T>(
+  v: T | undefined | null,
+  type: T
```
folex, line 49: So that's not "type", that's "default values", right?

src/lib/helpers/jsToAqua.ts
```diff
@@ -24,6 +24,43 @@ import { capitalize } from "./capitilize.js";
 import { cleanAquaName, validateAquaName } from "./downloadFile.js";
 import { stringifyUnknown } from "./jsonStringify.js";
 
+const OPTIONAL = "$$optional";
+const IS_NIL = "$$isNil";
```
folex, line 28: What are these? Comment would be nice

src/lib/helpers/jsToAqua.ts
```diff
@@ -24,6 +24,43 @@ import { capitalize } from "./capitilize.js";
...
+ * // v = { $$optional: 1, $$isNil: true }
+ * @example
+ * const v = makeOptional(null, 1);
+ * // v = { $$optional: 1, $$isNil: true }
+ */
+export const makeOptional = <T>(
+  v: T | undefined | null,
+  type: T
+): { [OPTIONAL]: T; [IS_NIL]?: boolean } => {
+  const isRepresentedAsNillInAqua = v === undefined || v === null;
```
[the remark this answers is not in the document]
folex: `null: string` or `undefined: u64` are "impossible" in Aqua type system.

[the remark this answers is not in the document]
folex: So, all in all, I think this logical statement is unsound:
```
const isRepresentedAsNillInAqua = v === undefined || v === null;
```

from value being either `undefined` or `null` it doesn't follow that it should be "represented as nil in Aqua".

src/lib/helpers/jsToAqua.ts
```diff
@@ -48,19 +85,22 @@ export const jsToAqua = (
   );
 
   return `${
-    typeDefs ?? ""
-  }\n\nfunc ${funcName}() -> ${type}:\n    <- ${value}\n`;
+    typeof typeDefs === "string" ? `${typeDefs}\n\n` : ""
```
folex, line 88: What... why?

When `typeDefs` are `string` and when not? Why silently ignore non-string `typeDefs`? Why event accept non-string `typeDefs`?

src/lib/helpers/jsToAqua.ts
```diff
@@ -152,13 +200,15 @@ export const jsToAquaImpl = (
       return error("All array elements must be of the same type");
     }
 
+    const ind = INDENTATION.repeat(newNestingLevel);
```
folex, line 203: I suggest to declare two similar variables:
```suggestion
    const newIndent = INDENTATION.repeat(newNestingLevel);
    const prevIndent = INDENTATION.repeat(nestingLevel);
```

And use them in the `value`. Right now, it is a variable in one place, in-place expression in another.

src/lib/helpers/jsToAqua.ts
```diff
@@ -152,13 +200,15 @@ export const jsToAquaImpl = (
...
+
     return {
       type: `[]${type}`,
-      value: `[${mappedToAqua
+      value: `[\n${ind}${mappedToAqua
         .map(({ value }) => {
           return value;
         })
-        .join(",")}]`,
+        .join(`,\n${ind}`)}\n${INDENTATION.repeat(nestingLevel)}]`,
```
folex, line 211:
```suggestion
        .join(`,\n${newIndent}`)}\n${prevIndent}]`,
```

src/lib/helpers/jsToAqua.ts
```diff
@@ -180,6 +230,25 @@ export const jsToAquaImpl = (
...
+      );
+
+      return {
+        type: `?${type}`,
+        value: IS_NIL in v && v[IS_NIL] === true ? NULL.value : `?[${value}]`,
+        typeDefs,
+      };
+    }
+
+    const newNestingLevel = nestingLevel === 0 ? 2 : nestingLevel + 1;
```
folex, line 249: DRY, especially around magic numbers

src/lib/helpers/jsToAqua.ts
```diff
@@ -204,15 +274,17 @@ export const jsToAquaImpl = (
...
         };
       },
       { keyTypes: [], keyDataTypes: [], entries: [] }
     );
 
     return {
       type: nestedType,
-      value: `${nestedType}(${entries.join(",")})`,
+      value: `${nestedType}(${entries.join(",")}\n${INDENTATION.repeat(
+        nestingLevel === 0 ? 1 : nestingLevel
```
folex, line 286: dry

test/__snapshots__/jsToAqua.test.ts.snap
```diff
@@ -0,0 +1,127 @@
...
+{
+  "type": "[]u64",
+  "typeDefs": undefined,
+  "value": "[
+        1,
+        2,
+        3
+]",
+}
+`;
```
folex, line 13: What are inputs, what are outputs? What is tested against what?

test/aqua/runDeployedWorkers.aqua
```diff
@@ -5,15 +5,15 @@ import "@fluencelabs/aqua-lib/builtin.aqua"
...
 
-export status
+export runDeployedServices
 
 data Answer:
     answer: string
     peer: string
 
-func status() -> *Answer:
+func runDeployedServices() -> *Answer:
```
folex, line 14: You have _the same_ code hardcoded in `src/lib/const.ts`. Any way to DRY them?

Would be very very nice to avoid hard-coding Aqua code as string in JS, because it is hard to change, compile and check hardcoded strings.

src/lib/const.ts
```diff
@@ -337,29 +337,29 @@ export const SEPARATOR = `\n\n${color.yellow(
...
-    print = (answer: string, peer: string):
-      Console.print([answer, peer])
+export runDeployedServices
 
-    answers: *string
+data Answer:
+    answer: string
+    peer: string
+
+func runDeployedServices() -> *Answer:
```
folex, line 350: You have the same code in `test/aqua/runDeployedWorkers.aqua`.

Is it possible to DRY that and keep only separate `.aqua` file?

test/jsToAqua.test.ts
```diff
@@ -14,58 +14,38 @@
...
+    ${true}      | ${"true"}   | ${"bool"}   | ${undefined}
+    ${false}     | ${"false"}  | ${"bool"}   | ${undefined}
+    ${null}      | ${"nil"}    | ${"?u8"}    | ${undefined}
+    ${undefined} | ${"nil"}    | ${"?u8"}    | ${undefined}
+    ${[]}        | ${"nil"}    | ${"?u8"}    | ${undefined}
+    ${{}}        | ${"nil"}    | ${"?u8"}    | ${undefined}
+    ${1}         | ${"1"}      | ${"u64"}    | ${undefined}
+    ${-1}        | ${"-1"}     | ${"i64"}    | ${undefined}
+    ${1.234}     | ${"1.234"}  | ${"f64"}    | ${undefined}
+    ${-1.234}    | ${"-1.234"} | ${"f64"}    | ${undefined}
```
folex, line 48: What about optional strings, numbers, etc?
```
"abc" '["abc"]' "?string"
```

test/jsToAqua.test.ts
```diff
@@ -80,6 +60,12 @@ describe("Conversion from js to aqua", () => {
     }
   );
 
+  test("Expect objects to match snapshot", () => {
+    obj.forEach((o) => {
+      expect(jsToAquaImpl(o, WRAPPER_FUNCTION_NAME, "")).toMatchSnapshot();
+    });
+  });
```
folex, line 67: No idea how to review these tests. Need you help on that.

test/jsToAqua.test.ts
```diff
@@ -14,58 +14,38 @@
  * limitations under the License.
  */
 
-import { jsToAquaImpl } from "../src/lib/helpers/jsToAqua.js";
+import { jsToAquaImpl, makeOptional } from "../src/lib/helpers/jsToAqua.js";
 
 const WRAPPER_FUNCTION_NAME = "res";
-const RETURN_TYPE = "Res";
 
 const obj = [
```
folex, line 21: What is it? How is it used?

test/__snapshots__/jsToAqua.test.ts.snap
```diff
@@ -0,0 +1,127 @@
...
+    d: ?u8
+    e: []ResE",
+  "value": "Res(
+        a=ResA(
+            b=[
+                2
+            ]
+        ),
+        c=nil,
+        d=nil,
```
folex, line 91: Let's make at least on of optionals non-nil.

src/lib/build.ts
shamsartem: Currently this function literally does nothing. I removed spinners because I tried a couple of them and they all just duplicate messages in some terminals, but a left the interface so in the future we can use some good spinner implementation that would work great

If the errror is thrown there - program will exit and spinner would have stopped together with that so there is nothing to worry about here

src/lib/const.ts
shamsartem: I just think it's mostly redundant, but I can return it back if you think it's not

src/lib/deployWorkers.ts
shamsartem: I tried to explain this concept to you previously. I will try once again. I created a function that infers aqua types from js/json values - so this is the reason I have to encode type in a value - it's all typesafe and pretty convenient to use. We could have created typescript types to aqua types tool but that's a whole can of worms. Regarding `Deals[string]` it's not generic so it's not type-parametrized. It's a Record with string keys and Deals as values so in order to get the deal type in typescript I access Deals record with any string. It's just a way to avoid creating a separate `Deal` type, but I can do it so it's more readoble to the people who don't write typescript whole day long

src/lib/deployWorkers.ts
shamsartem: It will not. In case there are no workers at all in the config and no workers were deployed before - it will put nil in aqua and create a type definition for DEFAULT_WORKER_NAME. Now that I am thinking about it again I am not completely sure that's the correct behaviour. I think I better not generate anyhing at all in this case instead of generating something, in a way, arbitrary. Thank you very much for your curiosity

src/lib/deployWorkers.ts
```diff
@@ -622,13 +633,89 @@ const validateWasmExist = async (
...
-  workersConfig: WorkersConfigReadonly
+  workersConfig: WorkersConfigReadonly,
+  fluenceConfig: FluenceConfigReadonly
 ) => {
+  const allWorkerNames = Object.keys(fluenceConfig.workers ?? {});
+
+  const workerNameDealsNills = allWorkerNames.reduce<
+    Record<string, ReturnType<typeof makeOptional<Deals[string]>>>
+  >((acc, workerName) => {
+    acc[workerName] = makeOptional(null, DEAL_TYPE);
```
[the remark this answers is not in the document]
shamsartem: I tried to simplify it. Please take a look after I push the commit

src/lib/helpers/jsToAqua.ts
shamsartem: In aqua standalone type definitions are required for any objects. This means that if the function infered some object - there will be some standalone type definitions. If there were some type defenitions generated - we want to insert them in to the file and separate them from other stuff with new lines. `typeDefs` is either string or undefined so by saing `typeof typeDefs === "string"` I mean `if there are any typeDefs`. e.g. if value is a scalar then there are no separate type definitions and typedefs === undefined

src/lib/helpers/jsToAqua.ts
shamsartem: added additional comment here

test/__snapshots__/jsToAqua.test.ts.snap
shamsartem: I improved snapshots and now you can see inputs and outps clearly in the snapshot itself

test/aqua/runDeployedWorkers.aqua
shamsartem: It's not exactly the same. This function is to check `worker deploy`

test/jsToAqua.test.ts
shamsartem: I added a couple of tests for optional scalars

test/jsToAqua.test.ts
shamsartem: No snapshots are better, you can check them out. What this is does is it generates a bunch of snapshots so you can look at the input and output and decide if they are correct

test/jsToAqua.test.ts
shamsartem: moved this inside tests so it's more clear how it's used

test/__snapshots__/jsToAqua.test.ts.snap
shamsartem: I reshuffeled test cases in such a way that it's now visible were there are nulls and when they happen and where are the optional values and when they happen including the case when optional is not null

src/lib/build.ts
folex: So you're relying on a global error handling to stop the CLI on any error.

If we ever break that invariant, will have to go and add `try finally` in a lot of places.

src/lib/deployWorkers.ts
```diff
@@ -659,65 +655,55 @@ export const ensureAquaFileWithWorkerInfo = async (
...
-            ? null
-            : workerHostsOptionals,
-          {
-            [DEFAULT_WORKER_NAME]: makeOptional(null, HOSTS_TYPE),
-          }
-        ),
-      },
-      "getWorkersInfo"
-    ),
+    jsToAqua(workersInfo, "getWorkersInfo"),
```
folex: nice

src/lib/helpers/jsToAqua.ts
```diff
@@ -24,14 +26,29 @@ import { capitalize } from "./capitilize.js";
...
 const OPTIONAL = "$$optional";
+/**
+ * When `$$isNil == true`, the value is represented as `nil` in aqua but it will still have type inferred from `$$optional` value.
+ */
 const IS_NIL = "$$isNil";
 
+type EmptyObject = Record<string, never>;
+
+/**
+ * Empty object and empty array are both considered nil in aqua
```
folex, line 42: Are they? I don't think `{} == nil` will yield `true` in Aqua 🤔

src/lib/helpers/jsToAqua.ts
```diff
@@ -85,11 +109,11 @@ export const jsToAqua = (
   );
 
   return `${
-    typeof typeDefs === "string" ? `${typeDefs}\n\n` : ""
+    typeDefs === undefined ? "" : `${typeDefs}\n\n`
```
folex, line 112: What can `typeDefs` be `undefined`?

src/lib/helpers/jsToAqua.ts
```diff
@@ -241,13 +258,13 @@ export const jsToAquaImpl = (
...
-        value: IS_NIL in v && v[IS_NIL] === true ? NULL.value : `?[${value}]`,
+        value: IS_NIL in v && v[IS_NIL] === true ? NIL.value : `?[${value}]`,
         typeDefs,
       };
     }
 
-    const newNestingLevel = nestingLevel === 0 ? 2 : nestingLevel + 1;
-    const ind = INDENTATION.repeat(newNestingLevel);
+    const newNestingLevel = nestingLevel + 1;
+    const newIndent = INDENTATION.repeat(newNestingLevel);
```
folex, line 267: you already had that defined a few lines above, right?

folex, line 267: you already had that defined a few lines above, right?
image

folex (approved):

src/lib/helpers/jsToAqua.ts
shamsartem: I fixed the description so it's more clear. There is just no {} in aqua at all

src/lib/deployWorkers.ts
```diff
@@ -659,65 +655,55 @@ export const ensureAquaFileWithWorkerInfo = async (
...
-    },
-    { ...workerNameDealsNills }
+  const directHostingWorkers = Object.fromEntries(
+    Object.entries({ ...fluenceConfig.workers, ...workersConfig.hosts }).map(
+      ([workerName, info]) => {
+        const key = workerName;
+        // if worker was deployed put hosts info, otherwise put null
+        const maybeHost = "relayId" in info ? info : null;
+        const value = makeOptional(maybeHost, HOSTS_TYPE);
+        return [key, value];
```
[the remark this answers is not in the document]
shamsartem: it can but currently it's set up in such a way that you will access it in one place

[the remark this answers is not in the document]
shamsartem: deployment won't change but access to what's been deployed will, because I think we will hardcode worker info in worker syntax instead of using arbitrary data structures

About deployment to several peers - this info is available still but under different name. `installation_spells` - it's an array. Completely unclear name, I agree. I will change it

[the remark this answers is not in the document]
folex: I'm completely lost :D

src/lib/helpers/jsToAqua.ts
shamsartem: these are typeDefs for objects. When it's a scalar - typeDefs are undefined - you check out tests to see that

src/lib/deployWorkers.ts
shamsartem: I just removed this name completely so it's not confusing and passed this directly to the function
(function has documentation and the argument to it is named properly so it's more clear what's that is)

PR #307, opened 2023-06-27
feat: use binary marine and mrepl releases for linux x86_64 and darwin x86_64 [fixes DXJ-121, DXJ-147]

shamsartem:


docs/commands/README.md
```diff
@@ -326,7 +326,7 @@ USAGE
   $ fluence dependency install [--no-input] [--force]
 
 FLAGS
-  --force     Force install even if the dependency/dependencies is/are already installed
+  --force     Force install using cargo even if the dependency/dependencies is/are already installed
```
folex, line 329: why cargo? `fluence dep i` is not cargo specific, right?

shamsartem: exactly. Thanks for catching that

src/lib/helpers/downloadFile.ts
```diff
@@ -68,8 +68,8 @@ export const downloadFile = async (
   }
 
   const arrayBuffer = await res.arrayBuffer();
-  await fsPromises.mkdir(dirname(path), { recursive: true });
-  await fsPromises.writeFile(path, new Uint8Array(arrayBuffer));
+  await mkdir(dirname(path), { recursive: true });
```
folex, line 71: what is `path`? I'd name it something like `outputPath`, so meaning is clear

what will happen is that directory already exists, will it fail or continue?

src/lib/helpers/downloadFile.ts
```diff
@@ -134,14 +134,14 @@ const downloadAndDecompress = async (
...
-    await fsPromises.access(dirPath);
+    await access(dirPath);
     return dirPath;
   } catch {}
 
   const archivePath = join(dirPath, ARCHIVE_FILE);
   await downloadFile(archivePath, get);
   await decompress(archivePath, dirPath);
-  await fsPromises.unlink(archivePath);
+  await rm(archivePath, { force: true, recursive: true });
```
folex, line 144: Why do you remove archive with `--rf`? It's super risky: a typo in a path, and you'll delete a whole directory.

Since you are removing a file, you do not need `recursive: true`.

src/lib/helpers/package.ts
```diff
@@ -249,7 +249,9 @@ export const handleInstallation = async ({
   version,
 }: HandleInstallationArg): Promise<void> => {
   const installAndMoveToDependencyPath = async (): Promise<void> => {
+    await rm(dependencyTmpPath, { recursive: true, force: true });
     await installDependency();
+    await rm(dependencyPath, { recursive: true, force: true });
```
folex, line 254: is it a dir? or a file? check if `--rf` is needed at all

src/lib/helpers/downloadFile.ts
shamsartem: `outputPath` would be better. directory that exists results in a rejection only when `recursive` is `false`.

src/lib/rust.ts
```diff
@@ -300,13 +305,36 @@ export const ensureCargoDependency = async ({
     });
 
   try {
-    if (process.env.CI === "true") {
+    if (
+      force ||
+      ![MARINE_CARGO_DEPENDENCY, MREPL_CARGO_DEPENDENCY].includes(name)
+    ) {
+      throw new Error("Install using cargo");
```
folex, line 312: What does this error mean? Install what, who?

src/lib/helpers/downloadFile.ts
shamsartem: agreed, no need for {recursive: true} here. will fix it. But operation is still safe cause all of this happens inside .fluence dir anyway

src/lib/rust.ts
```diff
@@ -300,13 +305,36 @@ export const ensureCargoDependency = async ({
...
+
+    try {
+      // check binary is working
+      const helpText = await execPromise({
+        command: join(dependencyPath, "bin", name),
+        args: ["--help"],
+      });
+
+      if (!helpText.includes(version)) {
+        throw new Error("Version mismatch");
```
folex, line 329: This error lacks a lot of information. I don't see how user/dev/me can address it.

What software has version mismatch? Linux? Chrome? What versions? Version 1 != version 2? What was expected, and what did you get?

src/lib/rust.ts
```diff
@@ -300,13 +305,36 @@ export const ensureCargoDependency = async ({
...
+      }
+    } catch {
+      await rm(dependencyPath, { recursive: true, force: true });
+      throw new Error("Not working. Fallback to normal installation");
```
folex, line 333: What is not working?

src/lib/rust.ts
```diff
@@ -300,13 +305,36 @@ export const ensureCargoDependency = async ({
...
   try {
-    if (process.env.CI === "true") {
+    if (
+      force ||
+      ![MARINE_CARGO_DEPENDENCY, MREPL_CARGO_DEPENDENCY].includes(name)
+    ) {
+      throw new Error("Install using cargo");
+    }
+
+    if (["darwin", "linux"].includes(platform()) && arch() === "x64") {
```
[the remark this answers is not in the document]
folex: Arch on my laptop is arm64, yet I'm able to use x86_64 binaries. Maybe check if `platform = darwin && arch = arm64` and download x86 anyway?

src/lib/helpers/package.ts
shamsartem: this is a dir. you can see a recreate it on the next line. Dirs do not remove without `recursive: true`. `force: true` on the other hand makes it so it doesn't fail if there is no directory to remove

folex: i'd name it as `dependencyDir` then, so it is clear

src/lib/rust.ts
```diff
@@ -241,7 +246,7 @@ const downloadPrebuiltCargoDependencies = async (
   try {
     await access(binaryPath);
   } catch {
-    const url = `https://github.com/fluencelabs/marine/releases/download/${name}-v${version}/${name}-linux-x86_64`;
+    const url = `https://github.com/fluencelabs/marine/releases/download/${name}-v${version}/${name}-${platform()}-x86_64`;
```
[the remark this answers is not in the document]
shamsartem: I do this check before using this function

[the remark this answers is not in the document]
folex: Then it would be better to pass platform here as an argument. After validation.

src/lib/rust.ts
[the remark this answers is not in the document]
shamsartem: it will not break anything because it will install from cargo like before. Anatoly doesn't build for arm64. I am not sure if it works good with rosetta. Can you please download x86_64 binary on your mac and check if it works?

src/lib/rust.ts
```diff
@@ -300,13 +305,36 @@ export const ensureCargoDependency = async ({
...
+      ![MARINE_CARGO_DEPENDENCY, MREPL_CARGO_DEPENDENCY].includes(name)
+    ) {
+      throw new Error("Install using cargo");
+    }
+
+    if (["darwin", "linux"].includes(platform()) && arch() === "x64") {
       await downloadPrebuiltCargoDependencies(name, version, dependencyPath);
     } else {
-      throw new Error("Not in CI");
+      throw new Error("Unsupported platform. Use cargo");
```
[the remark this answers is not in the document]
shamsartem: This error is not for the user. You can see it's catched right away and default cargo installation is used

src/lib/rust.ts
shamsartem: I understand your confusion, I will think about how can I avoid using these errors so it's more clear what is happening

PR #316, opened 2023-07-04
feat: print default dependencies and their versions when using `default` flag [fixes DXJ-398]

shamsartem:


boneyard93501 (approved):
awesome. thanks.

PR #317, opened 2023-07-04
feat: improve cli output when building project [fixes DXJ-406]

shamsartem:


src/lib/execPromise.ts
```diff
@@ -37,33 +41,40 @@ export type ExecPromiseArg = {
...
 
-  const commandToDisplay = IS_DEVELOPMENT
-    ? fullCommand
-    : fullCommand.replace(
-        /([\S\s]*--sk ).*([\S\s]*)/,
-        "$1<SECRET_KEY_IS_HIDDEN>$2"
-      );
+  const getCommandFailedMessage = (code: number | null = null) => {
+    return `Command: ${color.yellow(
+      IS_DEVELOPMENT
```
folex, line 60: How `IS_DEVELOPMENT` is set?

src/lib/execPromise.ts
```diff
@@ -37,33 +41,40 @@ export type ExecPromiseArg = {
...
+        ? fullCommand
+        : fullCommand.replace(
+            /([\S\s]*--sk ).*([\S\s]*)/,
+            "$1<SECRET_KEY_IS_HIDDEN>$2"
+          )
+    )} ${code === null ? "failed" : `exited with code ${code}`}`;
```
folex, line 66: how could `code` be null?

src/lib/execPromise.ts
```diff
@@ -73,13 +84,11 @@ export const execPromise = ({
...
 
-        rej(
-          new Error(
-            `Execution timed out: command didn't yield any result in ${color.yellow(
-              `${timeout}ms`
-            )}\n${debugInfo}`
-          )
-        );
+        res({
+          errorMessage: `${getCommandFailedMessage()}. Reason: Execution timed out: command didn't yield any result in ${color.yellow(
```
folex, line 88: > command didn't yield
what command? or will it be inside `getCommandFailedMessage`?

src/lib/execPromise.ts
```diff
@@ -126,32 +140,39 @@ export const execPromise = ({
...
     });
   });
+
+  if ("expectedErrorMessage" in result) {
+    throw new CLIError(result.expectedErrorMessage);
+  }
+
+  if ("errorMessage" in result) {
+    throw new Error(result.errorMessage);
+  }
```
folex, line 175: What is `expectedErrorMessage` and why is it different from `errorMessage`?

src/lib/execPromise.ts
```diff
@@ -126,32 +140,39 @@ export const execPromise = ({
...
+        res({
+          errorMessage: `${getCommandFailedMessage(code)}${getErrorMessage(
+            printOutput,
+            stderr
+          )}`,
+        });
       }
 
-      res(stdout);
+      res({ success: stdout });
```
folex, line 165: maybe name it `stdout`?

later you have `return result.success`, and it's pretty hard to guess that `success` means `stdout`.

src/lib/execPromise.ts
```diff
@@ -126,32 +140,39 @@ export const execPromise = ({
...
+
+  if ("expectedErrorMessage" in result) {
+    throw new CLIError(result.expectedErrorMessage);
+  }
+
+  if ("errorMessage" in result) {
+    throw new Error(result.errorMessage);
+  }
+
+  return result.success;
```
folex, line 177:
```suggestion
  return result.stdout;
```

src/lib/marineCli.ts
```diff
@@ -62,16 +63,28 @@ export const initMarineCli = async (
...
-    });
+    try {
+      return await execPromise({
+        command: marineCLIPath,
+        args,
+        flags,
+        spinnerMessage:
+          message === undefined
+            ? undefined
+            : getMessageWithKeyValuePairs(message, keyValuePairs),
```
folex, line 74: Let's move that to a separate variable plz? in-place takes 4 lines, making it hard to grasp

src/lib/marineCli.ts
```diff
@@ -62,16 +63,28 @@ export const initMarineCli = async (
...
+        options: { cwd },
+        printOutput,
+      });
+    } catch (e) {
+      if (
+        isObject(e) &&
+        hasKey("message", e) &&
+        typeof e["message"] === "string"
+      ) {
+        return commandObj.error(e["message"]);
```
folex, line 84: `errorMessage`, `expectedErrorMessage`, `message` – so many names for a conceptually same field 🤔

src/lib/rust.ts
```diff
@@ -285,9 +285,17 @@ const tryDownloadingBinary = async ({
 
   const url = `https://github.com/fluencelabs/marine/releases/download/${name}-v${version}/${name}-${platformToUse}-x86_64`;
 
+  startSpinner(
+    `Trying to download pre-built ${name}@${version} binary to ${replaceHomeDir(
```
folex, line 289:
```suggestion
    `Downloading pre-built ${name}@${version} binary to ${replaceHomeDir(
```

folex: I think "prebuilt" is kinda confusing here: user doesn't need to think about how our CI works

src/lib/rust.ts
```diff
@@ -285,9 +285,17 @@ const tryDownloadingBinary = async ({
...
+      dependencyDirPath
+    )}`
+  );
+
   try {
     await downloadFile(binaryPath, url);
+    stopSpinner();
   } catch {
+    stopSpinner("failed");
     return `Failed to download ${name}@${version} from ${url}`;
```
folex, line 299: Why no error in the message? `catch` should have caught some exception, right?

src/lib/execPromise.ts
```diff
@@ -126,32 +140,39 @@ export const execPromise = ({
...
-        );
+        res({
+          expectedErrorMessage: `\n${color.yellow(
+            MARINE_CARGO_DEPENDENCY
+          )} requires ${color.yellow(
+            "build-essential"
+          )} to be installed. Please install it and try again.\nOn debian-based systems (e.g. Ubuntu) you can install it using this command:\n\n${color.yellow(
+            "sudo apt install build-essential"
+          )}\n`,
+        });
```
folex, line 151: is that a structure? is that a multiline string? Very hard to follow all the `)} )} )}` on new lines.

Maybe create a string as a separate variable, and then do
```
let message = `\n${color.yellow(...`
res({expectedErrorMessage: message})
```

would be much easier to read
```

File: src/lib/execPromise.ts [hunk 12]
```diff
@@ -111,7 +120,12 @@ export const execPromise = ({
         stopSpinner(color.red("failed"));
       }
 
-      rej(new Error(`${printOutput ? "" : stderr}${error}\n${debugInfo}`));
+      res({
+        errorMessage: `${getCommandFailedMessage()}${getErrorMessage(
+          printOutput,
+          stderr
+        )}${error}`,
```
folex, line 127: same here. Suggest to move message to a separate variable to make it easier to read

src/lib/execPromise.ts
folex: ditto: suggest to move message to a separate variable to make it easier to read

src/lib/execPromise.ts
```diff
@@ -37,33 +41,40 @@ export type ExecPromiseArg = {
...
     startSpinner(spinnerMessage);
   }
 
-  const debugInfo = `Debug info:\n${commandToDisplay}\n`;
-
-  return new Promise<string>((res, rej): void => {
+  const result = await new Promise<
+    | { errorMessage: string }
+    | { expectedErrorMessage: string }
+    | { success: string }
```
folex, line 76:
```suggestion
    | { stdout: string }
```

src/lib/execPromise.ts
```diff
@@ -37,33 +41,40 @@ export type ExecPromiseArg = {
...
-export const execPromise = ({
+export const execPromise = async ({
   command,
   args,
   flags,
   options,
   spinnerMessage,
   timeout,
-  printOutput = false,
+  printOutput: printOutputArg = false,
```
folex, line 51: why `printOutputArg` starts from a lower case? looks strange

and why do you need explicit type here anyway?

src/lib/execPromise.ts
shamsartem: if you use ./bin/dev then it is development. Otherwise it's not
It's just so while developing you can see the actual args and in prod secret keys are not leaked
I think now that we don't use aqua cli this part can even be removed
I will do it

src/lib/execPromise.ts
shamsartem: typescript api says it can be null. I just handle this possibility. Don't know how to actually achieve this

src/lib/execPromise.ts
```diff
@@ -17,16 +17,20 @@
...
+import { isInteractive } from "./commandObj.js";
 import { IS_DEVELOPMENT, MARINE_CARGO_DEPENDENCY } from "./const.js";
 import { type Flags, flagsToArgs } from "./helpers/flagsToArgs.js";
 import { startSpinner, stopSpinner } from "./helpers/spinner.js";
 
 type SpawnParams = Parameters<typeof spawn>;
 
+const getErrorMessage = (printOutput: boolean, stderr: string) => {
+  return printOutput ? "" : `. Reason:\n${stderr}`;
+};
```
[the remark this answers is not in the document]
shamsartem: This shouldn't be resued anywhere in my opinion. This is to specific for this part of the codebase

src/lib/execPromise.ts
shamsartem: it's not a type it's a variable

src/lib/execPromise.ts
shamsartem: The command will be inside getCommandFailedMessage. I can move this to a variable. Not sure if it really impores readability at all

src/lib/marineCli.ts
```diff
@@ -62,16 +63,28 @@ export const initMarineCli = async (
...
+            : getMessageWithKeyValuePairs(message, keyValuePairs),
+        options: { cwd },
+        printOutput,
+      });
+    } catch (e) {
+      if (
+        isObject(e) &&
+        hasKey("message", e) &&
+        typeof e["message"] === "string"
+      ) {
```
folex, line 83: Maybe have a custom type for errors and do `instance of MyError`?

src/lib/execPromise.ts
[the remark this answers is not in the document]
folex: ok then

src/lib/execPromise.ts
shamsartem: expected error message (CLIError) is the same thing that happens when you do commandObj.error - it's just an error that is expected to be encountered by the user of the CLI. When expected error happens CLI doesn't print stack trace and also it is sent to Countly in a different way

src/lib/execPromise.ts
folex: I don't understand. What does `printOutput: printOutputArg = false` mean?

shamsartem: printOutput - is destructuring; printOutput: printOutputArg - is renaming of the destructured value; printOutput: printOutputArg = false - is a default value in case the destructured value is undefined

src/lib/execPromise.ts
folex: so that's basically a non-fatal error? I've seen languages separate Fatal error from others.

maybe name fields as `fatalError` and `nonFatalError`? not sure

src/lib/rust.ts
shamsartem: good catch. Will add

src/lib/execPromise.ts
shamsartem: they are all fatal, meaining that after they happen cli stops execution. The difference really is if the errors are expected or not. If they are expected then the error message is nice and user knows what to do next and we don't need to fix anything. But unexpected errors are something that should not normally happen

src/lib/execPromise.ts
folex: omg :) ok

src/lib/execPromise.ts
folex: okay. Let's at least document it in the code?

src/lib/marineCli.ts
shamsartem: It doesn't matter what is the type of the error here I think. if error has a message - then show it. The change is about treating marine errors as exepected errors so we don't show stack traces when marine command fails

src/lib/marineCli.ts
shamsartem: here is different. Error objects by convention have `message`. errorMessage and expectedErrorMessage is just a way to resolve promise and distinguish between the way it was resolved. I think we can avoid errorMessage, expectedErrorMessage and use instance of in execPromise instead, that would be better

src/lib/marineCli.ts
shamsartem: I imporoved that part, but this "message" will never change. This is what happens when you do new Error() in js - it always creates an object with a "message" property in it. CLIError extends default js error but is the same

folex: if you do `instanceof CLIError`, then you can access message as a field, via `.`, right?

src/lib/execPromise.ts
shamsartem: I imporved the code - check it out and tell me if any additional documentation is required

src/lib/marineCli.ts
shamsartem: It's true, but it doesn't matter what kind of error we have. Any error with a message is good. I added a genral check for that without using isntanceof, please check it out

src/lib/execPromise.ts
```diff
@@ -126,32 +136,31 @@ export const execPromise = ({
...
+        const commandFailedMessage = getCommandFailedMessage(code);
+        const errorMessage = getErrorMessage(printOutput, stderr);
+        res(new Error(`${commandFailedMessage}${errorMessage}`));
       }
 
       res(stdout);
     });
   });
+
+  if (result instanceof CLIError || result instanceof Error) {
```
folex, line 156: Aren't all `CLIError` are instances of `Error`?

src/lib/rust.ts
```diff
@@ -285,19 +286,19 @@ const tryDownloadingBinary = async ({
...
-  try {
-    await downloadFile(binaryPath, url);
-  } catch {
-    return `Failed to download ${name}@${version} from ${url}`;
-  }
+  startSpinner(
+    `Downloading ${name}@${version} binary to ${replaceHomeDir(
+      dependencyDirPath
+    )}`
+  );
```
folex, line 293:
```suggestion
  startSpinner(`Downloading ${name}@${version} binary to ${replaceHomeDir(dependencyDirPath)}`);
```

to me it's much easier to read, but if for you it's not, I'm ok with either

src/lib/rust.ts
```diff
@@ -285,19 +286,19 @@ const tryDownloadingBinary = async ({
...
 
   try {
-    await execPromise({
-      command: "chmod",
-      args: ["+x", binaryPath],
-    });
-  } catch {
-    return `Failed to make ${name}@${version} executable by running chmod +x '${binaryPath}'`;
+    await downloadFile(binaryPath, url);
+    stopSpinner();
```
folex, line 297: Maybe move `stopSpinner` to `finally`?

src/lib/typeHelpers.ts
```diff
@@ -106,3 +106,13 @@ export function getIsStringUnion<T extends string>(
     });
   };
 }
+
+export function isObjectWithStringMessage(
+  unknown: unknown
+): unknown is { message: string } {
```
folex, line 112: Why not `unknown is Error`?

folex (approved):

src/lib/execPromise.ts
shamsartem: Yeah. It is. Wanted to be overprotective in case CLIError implementation changes for some reason but seems like typescript would catch that issue for me. So it's good to remove additional check

src/lib/typeHelpers.ts
shamsartem: Because it's not true. I am not checking if it's an error here. That would be done with instanceof. I am thinking maybe you are right. I should just use instanceof. I am just inventing unnecessary generaliztions here

src/lib/rust.ts
shamsartem: the formatiing is done with prettier, no easy ways to control it and it's a good thing - less to argue about on code reviews

folex: yep, makes sense

src/lib/rust.ts
shamsartem: finally is for the cases when no matter error happened or not - you wanna do _the same thing_
here we do different things dependening on weateher error happened or not
if error happens - spinner is stopped with the message "fail"
if error does not happen - spinnner stops with the default message (in the spinner implementation that I previously used the message was "done")

so it looks like this

```
Downloading marine@0.1.1 binary to ~/.fluence/cargo/marine/0.1.1...failed
```

of


```
Downloading marine@0.1.1 binary to ~/.fluence/cargo/marine/0.1.1...done
```

but currently spinners are turned off

folex: ah, missed that. ok!

PR #349, opened 2023-07-13
feat: remove legacy deploy remains

shamsartem:


folex (approved):
Does it break backward compatibility with any of the existing examples?

PR #352, opened 2023-07-17
feat: Leave only workers.yaml, services.aqua, workers.aqua under version control in .fluence dir [fixes DXJ-412]

shamsartem:


src/lib/init.ts
```diff
@@ -124,16 +124,20 @@ export const init = async (options: InitArg = {}): Promise<FluenceConfig> => {
...
+
+    const dirHasOnlyGitInside =
+      directoryContent.length === 1 && directoryContent[0] === ".git";
+
+    if (isNonEmptyDir && !dirHasOnlyGitInside) {
+      return commandObj.error(
+        `Directory ${color.yellow(
+          projectPath,
+        )} is not empty. Please, init in an empty directory.`,
+      );
```
folex, line 139: wdyt about making it an option?

Like, if `--not-empty` passed, or a question `Directory is not empty, might override existing project. Still continue? Y/n` is answered `yes` – generate project in that directory.

folex (approved):

PR #353, opened 2023-07-17
chore: update air-beautify-wasm

shamsartem:


shamsartem:
renovate did it as well

PR #372, opened 2023-07-31
feat: improve config error messages [fixes DXJ-428]

shamsartem:


src/lib/configs/initConfig.ts
```diff
@@ -32,6 +32,32 @@ import { replaceHomeDir } from "../helpers/replaceHomeDir.js";
...
+  | null
+  | undefined;
+
+const validationErrorToString = (errors: AjvErrors) => {
+  if (errors === null || errors === undefined) {
+    return "";
+  }
+
+  return (
+    "Errors:\n\n" +
```
akim-bow, line 46: I suggest to use os.EOL instead hardcoding line separator

akim-bow (changes requested):

src/lib/configs/initConfig.ts
shamsartem: I started renaming yesterday but then I really wanted to know how are we doing on windows front in general at this moment so I installed a windows virtual machine to check this out. Even though there are still problems on windows which require additional development - new line characters does not seem to cause any problems for terminal messages

for files though, I wanna keep it consistent so the new line char is the same across all the systems so git doesn't freak out

So I think the best course of action would be to continue simply using `\n`

akim-bow (approved):

PR #374, opened 2023-08-01
feat: update dependencies

shamsartem:


src/lib/deployWorkers.ts
```diff
@@ -466,7 +460,9 @@ export const prepareForDeploy = async ({
     .filter(([workerName]) => {
       return workersToDeployConfirmed.includes(workerName);
     })
-    .map(([workerName, { peerIds = [] }]) => {
+    .map(([workerName, hostsOrDeals]) => {
+      const peerIds = "peerIds" in hostsOrDeals ? hostsOrDeals.peerIds : [];
+
```
akim-bow, line 465: What's the difference with the previous diff?

shamsartem: after I changed the types to be correct in this file above - now typescript complains, says there is no "peerIds" property in case of the deals. So we have to do it more explicitly so typescript doesn't complain

akim-bow (approved):

PR #380, opened 2023-08-03
feat: update aqua interfaces of the services that cli builds, improve config comments [fixes DXJ-435]

shamsartem:


src/lib/helpers/generateServiceInterface.ts
```diff
@@ -45,15 +50,97 @@ export const generateAquaInterfaceForService = async ({
...
+) => {
+  let previouslyGeneratedInterfacesStr = "";
+
+  try {
+    previouslyGeneratedInterfacesStr = await readFile(
+      await ensureFluenceAquaServicesPath(),
+      FS_OPTIONS,
     );
-  }
+  } catch {}
```
akim-bow, line 79: Lets add 1-2 tests on the new code?

src/lib/helpers/generateServiceInterface.ts
```diff
@@ -45,15 +50,97 @@ export const generateAquaInterfaceForService = async ({
...
+        return {
+          serviceId: getServiceIdFromServiceInterface(serviceDefinition),
+          serviceDefinition,
+        };
+      })
+      .filter(({ serviceId }) => {
+        return serviceNamesFromFluenceConfig.has(serviceId);
+      });
+
+  const setOfServiceIdsFromPreviouslyGeneratedInterfaces = new Set(
```
akim-bow, line 101: I don't like this naming. I think that typescript's job is to pass info about variable type and we needn't to duplicate type info into the variable name. Probably better naming is `serviceIdsFromPreviouslyGeneratedInterfaces`

akim-bow (approved):

PR #388, opened 2023-08-07
feat: lock CLI dependencies using `npm shrinkwrap` [fixes DXJ-441]

shamsartem:


shamsartem:
Cancel for now. Will try packing everything with yarn

PR #417, opened 2023-08-17
feat: use local npm

shamsartem:


src/types.d.ts
```diff
@@ -23,6 +23,11 @@ declare module "replace-homedir" {
   ): string;
 }
 
+declare module "node_modules-path" {
+  export = node_modules;
```
akim-bow, line 27: Is this export necessary?

PR #422, opened 2023-08-21
feat: patch oclif node execution so there are no warnings

shamsartem:


nahsi (approved):

akim-bow:
Maybe it's better to set env variable in package.json run script or somewhere else? This solution strongly depends on oclif source code and for me looks brittle.

src/beforeBuild.ts
```diff
@@ -97,3 +97,30 @@ await mkdir(COMPILED_INSTALLATION_SPELL_AQUA_WITH_TRACING_PATH, {
...
 await compileInstallationSpellAqua();
 await compileInstallationSpellAqua(true);
+
+const BIN_FILE_PATH = join(
+  join("node_modules", "oclif", "lib", "tarballs"),
+  "bin.js",
+);
+
+const binFileContent = await readFile(BIN_FILE_PATH, FS_OPTIONS);
+const NODE_RUN = `  "\\$NODE" "\\$DIR/run" "\\$@"`;
```
akim-bow, line 107: What if oclif will change this line in the next version?

shamsartem: Most likely the error will be thrown, as you can see from the code further down below

shamsartem:
> Maybe it's better to set env variable in package.json run script or somewhere else? This solution strongly depends on oclif source code and for me looks brittle.

there is no "run script in package.json". If you read this carefully you will see that this is a shell script which directly uses Node.js binary for executing js file. Because these shell scripts are a part of oclif's source code, which dictates the packaging process and the update mechanism (as I showed on the call) - I don't see a cleaner way to do this, then modifying it

PR #424, opened 2023-08-21
feat!: Remove old packages, add new ones

akim-bow:


shamsartem:
Is this a breaking change for CLI? My thinking is this should not be, I mean for the cli user not much is changing right?

shamsartem:
There should also be a small change in the `src/commands/dependency/versions.ts` and `src/beforeBuild.ts` where there is a copying of package.json of js-client which is required to display the js-client dependencies for the user when he executes `fluence dependency versions` command

shamsartem (approved):

nahsi (approved):

PR #434, opened 2023-08-28
fix: experimental warnings

shamsartem:


src/errorInterceptor.js
```diff
@@ -155,14 +155,13 @@ interceptor.on("response", ({ request: { url } }) => {
  */
 export function setUpProcessWarningListener() {
   process.on("warning", (warning) => {
-    if ("code" in warning && warning.code === "MODULE_NOT_FOUND") {
+    if (
+      ("code" in warning && warning.code === "MODULE_NOT_FOUND") ||
+      // when running in dev mode there is no warning.code
+      // so we have to rely on the text of the error message
+      (warning.stack ?? "").includes("Cannot find")
+    ) {
```
akim-bow, line 163: Are you sure that just silencing every warning is a good idea? Probably we need to log them somehow.
I suggest to throw error on already known warning and just log to console unknown. If someone encounter them, you will see it from user and will add additional condition for the next one.

shamsartem: Makes sence. Done

akim-bow (approved):

PR #436, opened 2023-08-28
feat: make docs and examples in the configs optional [fixes DXJ-458]

shamsartem:


.gitignore
```diff
@@ -9,7 +9,7 @@ oclif.manifest.json
 
 # for ease of development
 /schemas
-/.fluence-dev-project
+/.f
```
folex, line 12: huh?

src/lib/lifeCycle.ts
```diff
@@ -61,13 +62,49 @@ import {
...
 
   if (maybeUserConfig !== null) {
-    setUserConfig(maybeUserConfig);
+    const userConfig = maybeUserConfig;
+
+    if (userConfig.docsInConfigs === undefined) {
+      await handleDocsInConfigs(userConfig);
+    }
+
+    setUserConfig(userConfig);
```
folex, line 107: How `handleDocsInConfigs` is different from `setUserConfig`?

this code looks a little weird: you call `handleDocsInConfigs` and `setUserConfig` twice in the function. One time inside `if`, another time at the end.

Can you make it so it's called only a single time? So it reads easier, right now it's really hard to tell what will happen with all these IFs

src/lib/lifeCycle.ts
```diff
@@ -85,6 +122,12 @@ const ensureUserConfig = async (): Promise<void> => {
     await newUserConfig.$commit();
   }
 
+  commandObj.logToStderr(
+    `If you change your mind later, modify "countlyConsent" property in ${newUserConfig.$getPath()}`,
+  );
+
+  await handleDocsInConfigs(newUserConfig);
```
folex, line 129: What does `handleDocsInConfigs` do? Does it set config? Does it commit it to disk?

It doesn't return anything, yet it should have some effect, and then it is weird, because you do something implicit in `handleDocsInConfigs` and then something implicit in `setUserConfig`, and I guess their effects interdepends, and you can't call them in any order, right?

I'd suggest to
1. rename `handleDocsInConfigs` to something specific
2. make it return new value
3. use that value in `setUserConfig`

src/lib/lifeCycle.ts
shamsartem: 1. These are two completely different things. One "handles" docsInConfigs property in the configs - that is to say it determines weather it should be true or false. The other is setting this config as a global singelton (because it's very convenient to use and you can only have onle user config)
2. I can't do change so it is called only once because I am handlinngle two completely different possbilities: one is when there is no config at all. The other one is when there is a config but this property is missing in it (all our existing users fall into the latter cattegory)
3. I will try to improve the code

src/lib/lifeCycle.ts
```diff
@@ -160,7 +203,7 @@ export async function initCli<
   ]);
 
   await initCountly({ maybeFluenceConfig });
-  await handleFloxVersion(maybeFluenceConfig?.cliVersion);
+  await handleCliVersion(maybeFluenceConfig?.cliVersion);
```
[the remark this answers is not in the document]
shamsartem: agree. I will try to make it better.

src/lib/lifeCycle.ts
```diff
@@ -57,35 +60,85 @@ import { getIsInteractive } from "./helpers/getIsInteractive.js";
...
+      return commandObj.error("Unreachable");
+    },
+    onNoChoices: () => {
+      return commandObj.error("Unreachable");
+    },
+    options: [...DOCS_IN_CONFIGS_OPTIONS],
+  });
+
+  commandObj.logToStderr(
+    `If you change your mind later, modify "docsInConfigs" property in ${join(
```
boneyard93501, line 95: is that a global-only setting?

akim-bow (approved):

docs/configs/config.md
```diff
@@ -9,6 +9,7 @@ Defines global config for Fluence CLI
 | `countlyConsent`      | boolean                 | **Yes**  | Weather you consent to send usage data to Countly                                                                |
 | `version`             | number                  | **Yes**  |                                                                                                                  |
 | `dependencies`        | [object](#dependencies) | No       | (For advanced users) Global overrides of dependencies                                                            |
+| `docsInConfigs`       | boolean                 | No       | Whether to include commented-out documented config examples in the configs generated with the CLI                |
```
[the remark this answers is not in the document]
shamsartem: The link to the web docs will be in all of the configs by default

src/lib/lifeCycle.ts
shamsartem: this is per-user. I think it's fare. If cli user wants it to be different - he can go ahead and change it

PR #470, opened 2023-09-14
feat: add provider config [fixes DXJ-470]

shamsartem:


.github/actions/replace-version/dist/index.js
shamsartem, line 1: This is Anatoly's generated file for CI. Just so you don't accidentally start reviewing that

src/lib/configs/project/dockerCompose.ts
```diff
@@ -0,0 +1,315 @@
...
+        "--aqua-pool-size=5",
+        `-t=${tcpPort}`,
+        `-w=${webSocketPort}`,
+        "--external-maddrs",
+        `/dns4/${name}/tcp/${tcpPort}`,
+        `/dns4/${name}/tcp/${webSocketPort}/ws`,
+        "--allow-private-ips",
+        tcpPort === TCP_PORT_START
+          ? "--local"
+          : `--bootstraps=/dns/${bootStrapNoxName}/tcp/${TCP_PORT_START}`,
```
folex, line 196: Why some things are through ENV variables, and others are passed by flags?

I think it makes sense to pass as much information as possible via `--` flags. And everything else – via `Config.toml`

src/lib/configs/project/dockerCompose.ts
```diff
@@ -0,0 +1,315 @@
...
+      },
+      command: [
+        "--aqua-pool-size=5",
+        `-t=${tcpPort}`,
+        `-w=${webSocketPort}`,
+        "--external-maddrs",
+        `/dns4/${name}/tcp/${tcpPort}`,
+        `/dns4/${name}/tcp/${webSocketPort}/ws`,
+        "--allow-private-ips",
+        tcpPort === TCP_PORT_START
```
folex, line 194: What does this check mean?

src/lib/configs/project/dockerCompose.ts
folex: It's important to move as much as possible to Config.toml because Config.toml will be used in different networks-environments, while docker-compose is only for local.

So moving to Config.toml makes it solved for both public networks, and for local.

shamsartem: I am currently just doing the same stuff I hade in docker-compose in cli repo - I will imporve and extend all this later
You can suggest which porperties can be passed how better - would be useful

shamsartem: I am currently not generating Config.toml at all. I will move everything possible there in the next step

src/commands/aqua.ts
```diff
@@ -245,7 +246,7 @@ const resolveAbsoluteAquaPath = async ({
     return resolve(projectRootDir, maybePathFromFluenceYaml);
   }
 
-  const pathFromUserInput = await input(inputArg);
+  const pathFromUserInput = await input({ ...inputArg, default: cwd() });
```
folex, line 249: Can you please add a comment here on what this means?

I would expect default to be `projectRootDir` so that commands do not depend on the cwd, but apparently intention here is different.

src/commands/deal/deploy.ts
```diff
@@ -108,14 +113,17 @@ export default class Deploy extends BaseCommand<typeof Deploy> {
...
     await doRegisterIpfsClient(true);
     dbg("start running upload");
 
     const uploadResult = await upload(
       flags.tracing,
       uploadArg,
-      fluenceConfig[IPFS_ADDR_PROPERTY] ?? DEFAULT_IPFS_ADDRESS,
+      fluenceConfig[IPFS_ADDR_PROPERTY] ??
+        (fluenceEnv === "local"
+          ? FLUENCE_ENV_AQUA_IPFS_EXTERNAL_API_MULTIADDR
```
folex, line 125: Maybe call `FLUENCE_ENV_AQUA_IPFS_EXTERNAL_API_MULTIADDR` something like `LOCAL_IPFS_ADDRESS`?

src/commands/default/env.ts
```diff
@@ -46,7 +47,7 @@ export default class Peers extends BaseCommand<typeof Peers> {
       true,
     );
 
-    assert(envConfig !== undefined, "this command requires fluence project");
+    assert(envConfig !== null, "this command requires fluence project");
```
folex, line 50: sure there's no need to check for both null and undefined?

src/commands/default/env.ts
```diff
@@ -58,11 +59,12 @@ export default class Peers extends BaseCommand<typeof Peers> {
       newFluenceEnv === "custom" &&
       fluenceConfig.customFluenceEnv === undefined
     ) {
-      await ensureCustomRelays(fluenceConfig);
+      await ensureCustomNodes(fluenceConfig);
```
folex, line 62: I would avoid using name "Node"/"Nodes" to mean Nox instances. Better call them Peers or Relays.

Node is way too overloaded, and clashes with Node.js.

src/commands/default/env.ts
```diff
@@ -58,11 +59,12 @@ export default class Peers extends BaseCommand<typeof Peers> {
...
     }
 
     envConfig.fluenceEnv = newFluenceEnv;
     await envConfig.$commit();
+    await updateRelaysJSON({ fluenceConfig });
```
folex, line 67: besides, here you use `Relays`. I think `Relays` is the most precise name.

src/commands/provider/add-peer.ts
```diff
@@ -84,30 +65,26 @@ export default class AddPeer extends BaseCommand<typeof AddPeer> {
 
       const approveTx = await flt.approve(
         await matcher.getAddress(),
-        collateral * BigInt(workersCount),
+        collateral * BigInt(worker),
```
folex, line 68: basically it's now a Compute Unit count. Just FYI, no need to rename rn.

src/commands/default/env.ts
shamsartem: agree. Maybe rename this `@fluencelabs/fluence-network-environment` package - that's entirely the reason I call it like this - for consistency

src/commands/default/env.ts
shamsartem: with my ts super strict setup I can be completely sure about these things

src/lib/configs/project/dockerCompose.ts
```diff
@@ -0,0 +1,315 @@
...
+  providerConfig: ProviderConfigReadonly,
+): Promise<GetDefaultConfig> {
+  const def = await genDockerCompose(providerConfig);
+
+  return () => {
+    return yamlDiffPatch("", {}, def);
+  };
+}
+
+const migrations: Migrations<Config> = [];
```
folex, line 274: maybe move all constant and type declarations to the top of the file?

it's pretty hard to follow when functions are interleaved with consts and type defs

src/lib/configs/project/dockerCompose.ts
```diff
@@ -0,0 +1,315 @@
...
+    },
+    secrets: Object.fromEntries(
+      secretKeys.map(({ name, path: file }) => {
+        return [name, { file }];
+      }),
+    ),
+  };
+}
+
+async function genDefaultDockerCompose(
```
folex, line 264: Why is this function different from `initNewDockerComposeConfig`?

I would expect `initNewDockerComposeConfig` without parameters to generate default docker compose file.

But even their impl looks different.

src/lib/configs/project/dockerCompose.ts
```diff
@@ -0,0 +1,315 @@
...
+
+const initConfigOptions = {
+  allSchemas: [configSchemaV0],
+  latestSchema: configSchemaV0,
+  migrations,
+  name: DOCKER_COMPOSE_FILE_NAME,
+  getConfigOrConfigDirPath: getFluenceDir,
+};
+
+export async function initNewDockerComposeConfig(
```
folex, line 290: let's add some docs please, I don't understand the difference between `initNewDockerComposeConfig` and `initNewReadonlyDockerComposeConfig`

src/lib/configs/project/dockerCompose.ts
```diff
@@ -0,0 +1,315 @@
...
+): Promise<LatestConfig> {
+  const secretKeys = await genSecretKeysOrReturnExisting(
+    Object.keys(providerConfig.computePeers),
+  );
+
+  const [bootStrapNoxNameAndKey, ...restSecretKeys] = secretKeys;
+
+  assert(
+    bootStrapNoxNameAndKey !== undefined,
+    "At least one nox is required to generate a docker-compose.yml",
```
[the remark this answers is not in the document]
shamsartem: for local network you will need to have at least one nox. The fact that it will be bootstrap shouldn't matter. This error should never happen it's mostly a typescript pleaser because in TS there is no easiily available type "NonEmptyArray". I validate the config beforehand

src/lib/configs/project/dockerCompose.ts
```diff
@@ -0,0 +1,315 @@
...
+    await genDefaultDockerCompose(providerConfig),
+  )();
+}
+
+export const initDockerComposeConfig = getConfigInitFunction(initConfigOptions);
+
+export const initReadonlyDockerComposeConfig =
+  getReadonlyConfigInitFunction(initConfigOptions);
+
+export const dockerComposeSchema: JSONSchemaType<LatestConfig> = configSchemaV0;
```
folex, line 315: docs here would be nice. rn the purpose of these consts is unclear to me.

src/lib/configs/project/fluence.ts
```diff
@@ -496,6 +501,39 @@ const configSchemaV3: JSONSchemaType<ConfigV3> = {
   },
 } as const;
 
+const configSchemaV3: JSONSchemaType<ConfigV3> = configSchemaV3Obj;
```
folex, line 504: There's 400 (!!!) lines between `configSchemaV0` and `configSchemaV3` declarations. That's wild!

Let's group consts at the top, then types, then functions.

Also I suggest to separate this file into several, it's way too big.

src/lib/configs/project/fluence.ts
```diff
@@ -506,11 +544,11 @@ const getDefaultConfig = async (): Promise<string> => {
 # You can use \`fluence init\` command to generate a template for new Fluence project
```
folex, line 544: suggest to move that text to a separate file

src/lib/configs/project/provider.ts
```diff
@@ -0,0 +1,202 @@
...
+export type ProviderConfig = InitializedConfig<LatestConfig>;
+export type ProviderConfigReadonly = InitializedReadonlyConfig<LatestConfig>;
+
+const validate: ConfigValidateFunction<LatestConfig> = (config) => {
+  const invalid: Array<{
+    offerName: string;
+    missingComputePeerNames: Array<string>;
+  }> = [];
+
+  for (const [offerName, { computePeers }] of Object.entries(config.offers)) {
```
folex, line 151: some doc here would be nice

src/commands/aqua.ts
shamsartem: by default input for aqua is set in fluence.yaml. If it's not set (you removed it for some reason from fluence.yaml) or you don't have fluence.yaml at all - then there is need to provide path to the input for aqua compilation. You can provide path using a flag or you will be prompted and making cwd default in this case seems reasonable. But you can still input path that you need by hand. This behaviour didn't change for a while, I am just changing the way default's are displayed/handled in all prompts.

src/lib/const.ts
```diff
@@ -178,11 +201,11 @@ export const AUTO_GENERATED = "auto-generated";
...
 export const DEFAULT_WORKER_NAME = "defaultWorker";
 
-const KEY_PAIR_FLAG_NAME = "key-pair-name";
+const SK_FLAG_NAME = "sk";
 export const KEY_PAIR_FLAG = {
-  [KEY_PAIR_FLAG_NAME]: Flags.string({
+  [SK_FLAG_NAME]: Flags.string({
     char: "k",
-    description: "Key pair name",
+    description: "Peer secret key name",
```
folex, line 208:
```suggestion
    description: "Name of a peer's Network Private Key",
```

src/commands/default/env.ts
shamsartem: I will call this type and these functions `AddrAndPeerId` cause that's what these object are actually

src/lib/deployWorkers.ts
```diff
@@ -488,126 +488,131 @@ export const prepareForDeploy = async ({
...
-              ...overridesFromProject,
-            };
+            const { [FACADE_MODULE_NAME]: facadeModule, ...restModules } =
+              serviceConfig.modules;
+
+            const modules = [
+              ...Object.entries(restModules),
+              [FACADE_MODULE_NAME, facadeModule] as const,
+            ].map(([name, { get, ...overridesFromService }]) => {
+              const moduleUrlOrAbsolutePath = getUrlOrAbsolutePath(
```
folex, line 541: maybe move that to a separate function? indentation level is pretty big here.

src/lib/generateUserProviderConfig.ts
```diff
@@ -0,0 +1,159 @@
...
+  const userProvidedConfig: UserProvidedConfig = {
+    computePeers: {},
+    offers: {},
+  };
+
+  let isAddingMoreComputePeers: boolean;
+  let computePeersCounter = 0;
+
+  do {
+    const defaultName = `nox-${computePeersCounter}`;
```
folex, line 61: move to a separate function, so that control flow is more visible?

src/lib/configs/project/dockerCompose.ts
shamsartem: I have this in all configs. It's set up like that so the types work when config is extended. `JSONSchemaType<LatestConfig>` is not exactly the same as a `configSchemaV0 as const` type.

src/lib/configs/project/fluence.ts
shamsartem: I really would suggest configs refactoring to a separate PR

src/lib/configs/project/dockerCompose.ts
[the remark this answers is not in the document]
folex: it's just strange that you check against `bootStrapNoxNameAndKey`, not against number of noxes user requested

src/commands/aqua.ts
folex: can you explain that in a comment to this line please?

src/commands/default/env.ts
folex: I think `PeerAddress` is easier to grasp. `and` in names is hard.

src/lib/configs/project/dockerCompose.ts
folex: Let's document that in comments, so that intentions are clear.

src/lib/configs/project/dockerCompose.ts
folex: let's document that in code plz

docs/configs/provider.md
```diff
@@ -22,10 +22,11 @@ Defines config used for provider set up
...
-| `port`   | string | No       | Both host and container port to use. Default: for each nox a unique port is assigned starting from 9991 |
-| `worker` | number | No       |                                                                                                         |
+| Property        | Type   | Required | Description                                                                                                       |
+|-----------------|--------|----------|-------------------------------------------------------------------------------------------------------------------|
+| `tcpPort`       | number | No       | Both host and container TCP port to use. Default: for each nox a unique port is assigned starting from 7771       |
+| `webSocketPort` | number | No       | Both host and container WebSocket port to use. Default: for each nox a unique port is assigned starting from 9991 |
+| `worker`        | number | No       |                                                                                                                   |
```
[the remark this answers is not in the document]
folex: also, probably rename `worker` to `computeUnits`

src/lib/multiaddres.ts
```diff
@@ -288,7 +294,8 @@ export async function resolveRelay({
   ...args
 }: { maybeRelay: string | undefined } & ResolveNodesArgs) {
   return (
-    (await getMaybeNamedNode(maybeRelay, args.maybeFluenceConfig))?.multiaddr ??
+    (await getMaybeNamedAddrAndPeerId(maybeRelay, args.maybeFluenceConfig))
+      ?.multiaddr ??
     maybeRelay ??
     (await getRandomRelayAddr(args))
   );
```
folex, line 301: this is super hard to follow. Two ternary operators is too much. Especially with inlined functions.

I'd suggest to move all the function calls to separate variables, and then maybe use `if`s

src/lib/configs/project/dockerCompose.ts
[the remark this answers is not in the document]
shamsartem: It's because of the typescript. I have to convince it I checked the first element in the array actually exists. It can't infer this idea from just checking the array length before that

src/commands/aqua.ts
shamsartem: sure

src/commands/default/env.ts
shamsartem: it's a bit harder but that's literally what we have here. Maybe we shouldn't have it at all? Always have just a simple array of multiaddrs and a simple function to get peerId from multiaddr? js-client users would rarely need peerId right? Maybe we expose only multiaddrs?

src/lib/configs/project/dockerCompose.ts
shamsartem: I will do that in a spearate PR when refactoring configs. I value the suggestion and will imporve this

src/commands/default/env.ts
folex: js-client always needs peer id in the multiaddr. I guess that's why there's a separate peer_id field in there.

but yeah, just multiaddr would be much simpler. it just would take more validation.

src/lib/configs/project/dockerCompose.ts
shamsartem: I will. It's natural to code until it's clear that what you are doing is pretty hard and involved and then to refactor and document. It would be a waste if I did it from the begining  https://linear.app/fluence/issue/DXJ-509/refactor-configs-in-cli

docs/configs/provider.md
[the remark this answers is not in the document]
shamsartem: Can rename. Also could you please help me with a nice description? Can't think of anything good. Also as you might notice I don't indent this in a separate group as you did it in your take in notion. I think it's deep enough, probably would be better to just improve the names like in here

src/lib/multiaddres.ts
shamsartem: ok. will refactor. but thease are not ternaries

src/lib/multiaddres.ts
```diff
@@ -361,23 +368,22 @@ export async function updateRelaysJSON({
...
+    Object.entries(providerConfig.computePeers).map(async ([name, peer], i) => {
+      const { webSocketPort: webSocketPortFromConfig, worker = 1 } = peer;
 
-        if (port === undefined) {
-          port = `${WEB_SOCKET_PORT_START + i}`;
-        }
+      const webSocketPort =
+        webSocketPortFromConfig === undefined
+          ? WEB_SOCKET_PORT_START + i
+          : webSocketPortFromConfig;
```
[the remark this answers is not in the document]
shamsartem: I will use `??` here so it will be a lot nicer

src/commands/default/env.ts
shamsartem: But the problem is that js-client accepts mutiadr as a string anyway and validates it. So there is no much reason to have peer_id field in there. It's totally derivable from address alone

src/lib/deployWorkers.ts
shamsartem: Moved it to a separate function. Not sure if I made it a lot better

src/commands/aqua.ts
```diff
@@ -245,7 +246,12 @@ const resolveAbsoluteAquaPath = async ({
     return resolve(projectRootDir, maybePathFromFluenceYaml);
   }
 
-  const pathFromUserInput = await input(inputArg);
+  // By default input for aqua is set in fluence.yaml
+  // If it's not set (user removed it for some reason from fluence.yaml)
+  // or there is no fluence.yaml at all - then there is a need to provide path to the input for aqua compilation
+  // By default it's suggest to use current working directory as input for aqua compilation
```
akim-bow, line 252:
```suggestion
  // By default it's suggested to use current working directory as input for aqua compilation
```

akim-bow:
Can you explain the motivation for having **fluence local** commands which just calls same commands from the **docker compose**? Wouldn't it be easier for user to call familiar docker compose commands directly instead of learning additional commands of more specific tool?

src/lib/const.ts
```diff
@@ -36,10 +36,26 @@ export const RUST_WASM32_WASI_TARGET = "wasm32-wasi";
...
+export const numberProperties = [
+  "minPricePerWorkerEpoch",
+  "maxCollateralPerWorker",
+] as const;
+
+export type NumberProperty = (typeof numberProperties)[number];
+
+export const defaultNumberProperties: Record<NumberProperty, number> = {
+  maxCollateralPerWorker: 1 * 10 ** 18,
+  minPricePerWorkerEpoch: 83 * 10 ** 15,
```
akim-bow, line 48: I suggest to extract these values to constant variable.
For example:
const ETH = 10 ** 18;
const MILLI_ETH = 10 ** 15;
...
{
maxCollateralPerWorker: ETH,
minPricePerWorkerEpoch: 83 * MILLI_ETH,
}

src/lib/generateUserProviderConfig.ts
```diff
@@ -0,0 +1,202 @@
...
+
+export async function generateUserProviderConfig({
+  numberOfNoxes,
+}: ProviderConfigArgs) {
+  const userProvidedConfig: UserProvidedConfig = {
+    computePeers: {},
+    offers: {},
+  };
+
+  let isAddingMoreComputePeers = true;
```
akim-bow, line 57:
```suggestion
  let hasMoreComputePeers = true;
```

src/lib/generateUserProviderConfig.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  isAddingMoreComputePeers =
+    numberOfNoxes === undefined && isInteractive
+      ? await confirm({
+          message: "Do you want to add more compute peers",
+        })
+      : (numberOfNoxes ?? DEFAULT_NUMBER_OF_NOXES) > computePeersCounter;
+
+  return { computePeersCounter, isAddingMoreComputePeers };
+}
```
akim-bow, line 129:
```suggestion
async function addComputePeer({
  numberOfNoxes,
  userProvidedConfig,
}: AddComputePeerArgs) {
  let computePeersCounter = 0;
  let isAddingMoreComputePeers = true;
  do {  
  const defaultName = `nox-${computePeersCounter}`;

  let name =
    numberOfNoxes === undefined
      ? await input({
          message: `Enter name for compute peer`,
          default: defaultName,
        })
      : defaultName;

  if (name === defaultName) {
    name = defaultName;
    computePeersCounter = computePeersCounter + 1;
  }

  const slotsStr = await input({
    message: `Enter number of workers for ${color.yellow(name)}`,
    default: "1",
    validate: validatePositiveNumberOrEmpty,
  });

  userProvidedConfig.computePeers[name] = {
    computeUnits: Number(slotsStr),
  };

  isAddingMoreComputePeers =
    numberOfNoxes === undefined && isInteractive
      ? await confirm({
          message: "Do you want to add more compute peers",
        })
      : (numberOfNoxes ?? DEFAULT_NUMBER_OF_NOXES) > computePeersCounter;
      } while (isAddingMoreComputePeers);
}
```

src/lib/generateUserProviderConfig.ts
```diff
@@ -0,0 +1,202 @@
...
+  let computePeersCounter = 0;
+
+  do {
+    ({ computePeersCounter, isAddingMoreComputePeers } = await addComputePeer({
+      computePeersCounter,
+      numberOfNoxes,
+      userProvidedConfig,
+      isAddingMoreComputePeers,
+    }));
+  } while (isAddingMoreComputePeers);
```
akim-bow, line 67:
```suggestion
addComputePeer({
      numberOfNoxes,
      userProvidedConfig,
    }));
```

src/lib/generateUserProviderConfig.ts
```diff
@@ -0,0 +1,202 @@
...
+  offer[property] = Number(propertyStr);
+}
+
+const DEFAULT_NUMBER_OF_NOXES = 3;
+
+export type ProviderConfigArgs = {
+  numberOfNoxes?: number | undefined;
+};
+
+export async function generateUserProviderConfig({
```
akim-bow, line 49: I suggest to refactor body of this function. See above.
This refactoring will make this function cleaner, moving implementation details in more specific functions which facilitate a level of abstraction.

If you agree please apply this refactoring to all `do... while` constructions

src/lib/helpers/package.ts
```diff
@@ -349,7 +349,9 @@ export const getRecommendedDependencies = (packageManager: PackageManager) => {
 
   return Object.fromEntries(
     Object.entries(versionsPerPackageManager).filter(([dependencyName]) => {
-      return defaultDependencies.includes(dependencyName);
+      return defaultDependencies.some((dep) => {
+        return dep === dependencyName;
+      });
```
akim-bow, line 354: Please explain this change. It looks same for me.

src/lib/helpers/validations.ts
```diff
@@ -85,3 +85,23 @@ export const validateAllVersionsAreExact = async (
...
+
+  if (Number.isNaN(parsed)) {
+    return "Must be a number";
+  }
+
+  if (parsed <= 0) {
+    return "Must be a positive number";
+  }
+
+  return true;
```
akim-bow, line 106:
```suggestion
if (parsed > 0) {
  return true;
}

return "Must be a positive number";
```

Comparing to NaN always gives `false`

src/lib/multiaddres.ts
```diff
@@ -279,33 +289,37 @@ function getRandomRelayAddr(
...
+
+  if (namedAddr !== undefined) {
+    return namedAddr;
+  }
+
+  if (maybeRelay !== undefined) {
+    return maybeRelay;
+  }
+
+  return getRandomRelayAddr(args);
```
akim-bow, line 308:
```suggestion
return namedAddr ?? maybeRelay ?? getRandomRelayAddr(args);
```

src/lib/multiaddres.ts
```diff
@@ -319,3 +333,62 @@ export function getPeerId(addr: string): string {
...
+
+export async function updateRelaysJSON({
+  fluenceConfig,
+  numberOfNoxes,
+}: UpdateRelaysJSONArgs) {
+  if (
+    !(
+      typeof fluenceConfig?.relaysPath === "string" &&
+      envConfig?.fluenceEnv !== undefined
+    )
```
akim-bow, line 350:
```suggestion
    typeof fluenceConfig?.relaysPath !== "string" ||
      envConfig?.fluenceEnv === undefined
```

Feels a bit simpler

shamsartem:
> **fluence local** commands which just calls same commands from the **docker compose**? Wouldn't it be easier for user to call familiar docker compose commands directly instead of learning additional commands of more specific tool?

@akim-bow probably yeah. Hard to say. But @boneyard93501 wanted to have some wrapper for this. I am still not 100% certain we should do this. This is better only for super beginners who don't know what they are doing. If a person know docker and stuff - he can still use it directly cause the config is there

src/lib/generateUserProviderConfig.ts
shamsartem: sounds ok

src/lib/multiaddres.ts
shamsartem: yeah I just recently changed that so it's not so much nullish stuff, because folex asked me to. But if I move away the await outside like I did - nullish stuff doesn't look that bad anymore, that's true

src/lib/helpers/package.ts
shamsartem: I don't actually know the reason why ts started to fail here when generating docs. Previously it was fine

shamsartem: https://linear.app/fluence/issue/DXJ-510/investigate-why-ts-error-happens-when-generating-docs

akim-bow (approved):

PR #478, opened 2023-09-20
feat: add `showSubnet` func

shamsartem:


folex (approved):

folex:
would be nice to add a test for that function

PR #480, opened 2023-09-21
feat: workers remove

shamsartem:


src/commands/workers/remove.ts
```diff
@@ -0,0 +1,169 @@
...
+      return commandObj.error(
+        `There are no workers in ${FLUENCE_CONFIG_FULL_FILE_NAME}`,
+      );
+    }
+
+    const workerNamesSet = Object.keys(workersConfig.hosts).map(
+      (workerName) => {
+        return workerName;
+      },
+    );
```
akim-bow, line 83: Looks like you can safely remove this `.map`

shamsartem: oh ahaha. that's for sure. thanks!

akim-bow (approved):

PR #484, opened 2023-09-22
fix: npm not being found when installed with yarn [fixes DXJ-479]

shamsartem:


src/lib/npm.ts
```diff
@@ -38,20 +39,50 @@ const getNpmPath = async () => {
   return join(node_modules(), ".bin", "npm");
 };
 
+async function runNpm(args: Omit<ExecPromiseArg, "command">) {
+  let res: string;
+
+  try {
+    res = await execPromise({
```
folex, line 46: maybe first check that npmPath exists and is an executable?

and then use the correct path. could even cache it.

src/lib/npm.ts
```diff
@@ -39,32 +40,22 @@ const getNpmPath = async () => {
...
-      commandObj.error(
-        `Can't find ${color.yellow(
-          "npm",
-        )} in your system. Please make sure it's available on PATH. Errors:\n\n${stringifyUnknown(
-          localNpmErr,
-        )}\n\n${stringifyUnknown(globalNpmErr)}`,
-      );
-    }
+    const npmPath = await getNpmPath();
+    await access(npmPath);
```
folex, line 52: How will the error look like? Maybe catch the error, and explain what happened? Which paths were tried, which ENVs, and so on.

shamsartem: I improved the error message so now it will look like this

```
Error: Not able to install @fluencelabs/registry@0.8.7 to /home/shams/.fluence/npm/@fluencelabs/registry/0.8.7. 
Error: Couldn't find npm executable. Tried at /home/shams/Projects/work/fluence-cli/node_modules/.bin/npm and also tried finding npm in your $PATH. Please make sure npm is available
```

src/lib/npm.ts
```diff
@@ -38,20 +40,40 @@ const getNpmPath = async () => {
...
+  if (typeof npmExecutable === "string") {
+    return execPromise({ command: npmExecutable, ...args });
+  }
+
+  try {
+    const npmPath = await getNpmPath();
+    await access(npmPath);
+    npmExecutable = npmPath;
+  } catch {
+    npmExecutable = "npm";
```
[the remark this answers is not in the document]
shamsartem: in js you can't do a second `catch` like this and there is no good way to do `resolveBinaryFromPATH` cause all this execution is done by nodejs by spwaning a default shell in your system. So I decided to just run `npm --version` - if this command fails - then we can pretty safely assume that npm is not installed and can print an error that will show everything that we tried

akim-bow (approved):

PR #486, opened 2023-09-22
feat: add config formatting

shamsartem:


src/lib/configs/initConfig.ts
```diff
@@ -438,6 +438,58 @@ export function getReadonlyConfigInitFunction<
...
+    .trim()
+    .split("\n")
+    .flatMap((line, i, ar) => {
+      // If it's empty string - it was a newline - remove it
+      if (line.trim() === "") {
+        return [];
+      }
+
+      const maybePreviousLine = ar[i - 1];
+
```
akim-bow, line 452: Can we check `undefined` once at a top of the function?

src/lib/configs/initConfig.ts
```diff
@@ -438,6 +438,58 @@ export function getReadonlyConfigInitFunction<
...
+  return `${formattedConfig.trim()}\n`;
+}
+
+async function saveConfig(
+  configPath: string,
+  migratedConfigString: string,
+): Promise<string> {
+  const configToSave = formatConfig(migratedConfigString);
+  await writeFile(configPath, configToSave, FS_OPTIONS);
+  const savedConfig = configToSave;
```
akim-bow, line 489:
```suggestion
return configToSave;
```

akim-bow: I don't see a point to make another variable like this.
There is already variable named **configToSave** which stored config. It guess it's pretty obvious what's saving here.

shamsartem: Wanted to be a bit more explicit, but looks a little bit questionable, I can remove

src/lib/configs/initConfig.ts
shamsartem: I reworked this part so no it's less comments - more veriable names instead and one less if statement - but the result is the same

akim-bow (approved):

PR #502, opened 2023-09-28
feat!: improve fluence env switching [fixes DXJ-473]

shamsartem:


src/lib/init.ts
```diff
@@ -58,19 +56,94 @@ import {
...
+const main = async () => {
+  await Fluence.connect(connectTo);
+
+  const helloWorldResult = await helloWorld("Fluence");
+  const helloWorldRemoteResult = await helloWorldRemote("Fluence");
+  const getInfoResult = await getInfo();
+
+  console.log(helloWorldResult);
+
+  process.exit(0);
```
akim-bow, line 139: why not `await Fluence.disconnect()` ?

nahsi (approved):

src/lib/multiaddres.ts
```diff
@@ -27,89 +27,234 @@ import { multiaddr } from "@multiformats/multiaddr";
...
+}
 
-const getRandomArrayItem = <T>(ar: Array<T>): T => {
+function getRandomArrayItem<T>(ar: Array<T>): T {
   const randomIndex = Math.round(Math.random() * (ar.length - 1));
   const randomItem = ar[randomIndex];
   assert(randomItem !== undefined);
   return randomItem;
-};
+}
```
akim-bow, line 205: I think this util function is too smart and specific.
Probably you can have a function for getting random integer from range and use that in your code.

src/commands/module/remove.ts
```diff
@@ -115,6 +115,8 @@ export default class Remove extends BaseCommand<typeof Remove> {
       serviceConfig,
     );
 
+    // Our configs can't be Map object so there is no workaround for this rule
+    // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
     delete serviceConfig.modules[moduleNameToRemove];
```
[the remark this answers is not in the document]
shamsartem: I agree it can be dealt with. Even created a task https://linear.app/fluence/issue/DXJ-484/use-js-map-objects-instead-of-plain-js-object-for-dynamic-hashmaps-in
This would very rarelly cause a problem but it would require time to fix this, so if you don't mind I will do it later in a separate PR

src/commands/run.ts
```diff
@@ -212,12 +212,10 @@ export default class Run extends BaseCommand<typeof Run> {
...
-    if (stringResult !== undefined) {
-      // If `--quite` flag is used then commandObj.log does nothing
-      // So we use console.log here instead
-      // eslint-disable-next-line no-console
-      console.log(stringResult);
-    }
+    // If `--quite` flag is used then commandObj.log does nothing
+    // So we use console.log here instead
+    // eslint-disable-next-line no-console
+    console.log(stringResult);
```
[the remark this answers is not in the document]
shamsartem: that's a good question. --quite means "don't clutter output with cli logs". If you print --help you will see the flag description

"Print only execution result. Overrides all --log-level-* flags"

I am not sure if there is time when user would absolutely require for some reason to not see the result of aqua execution

src/lib/configs/initConfig.ts
```diff
@@ -556,6 +564,7 @@ export function getConfigInitFunction<
             Object.prototype.hasOwnProperty.call(config, key) &&
             typeof config[key] === "function"
           ) {
+            // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
             delete config[key];
```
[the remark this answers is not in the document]
shamsartem: interesting trick. I will try

src/lib/configs/project/fluence.ts
```diff
@@ -853,6 +885,22 @@ const validateHostsAndDeals = (
...
+
+  const { relays } = customFluenceEnv;
+
+  if (relays.length === 0) {
+    return "customFluenceEnv.relays list must not be empty";
+  }
+
+  return true;
+}
+
```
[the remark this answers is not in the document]
shamsartem: There is some custom validation there that is not achievable by merely using a validation lib
The `yup` kind of validation is done using ajv
But it's a good point. I totally forgot ajv has a "minItems" check for arrays - used that intead of this custom validation

src/lib/init.ts
shamsartem: for this example it prefectly works without it. But wouldn't harm. Will add

shamsartem: I tried refactoring it a bit - check it out after I push

src/lib/multiaddres.ts
shamsartem: don't really understand what you are suggesting here. This function seems pretty small and to-the-point - it just takes a random item from array, don't really know how to decompose it further, if that's what you want

src/commands/run.ts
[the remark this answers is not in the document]
akim-bow: Up to you to decided i think. If you think this is intentional, you can close it

src/lib/multiaddres.ts
akim-bow: Yes, the initial comment was about some decomposition. Add an additional function for getting random integer in rand and rewrite it in your source code.

```
const r = randInt(0, array.lenght);
const el = arr[r] || null;
```

But can we use instead of this utility function just `lodash.sample` ?

shamsartem: yes. Just using lodash seems to be fine, I think I didn't have lodash as a dependency in the project when I wrote this function

akim-bow:
Thanks, GJ!

akim-bow (approved):

PR #517, opened 2023-10-03
feat: use aqua-to-js lib [fixes DXJ-489]

shamsartem:


src/lib/aqua.ts
```diff
@@ -52,27 +52,30 @@ const getAquaFilesRecursively = async (dirPath: string): Promise<string[]> => {
 
 const writeFileAndMakeSureDirExists = async (
   filePath: string,
-  data: string,
+  dataPromise: string | Promise<string | undefined>,
```
akim-bow, line 55: this makes function too smart - it executes _any_ effectful callback. This function could operate just on `string | undefined` type.
```suggestion
  fileContent: string | undefined,
```

akim-bow: In the code below, in flatMap callback, you can await result before passing it to this function.

src/lib/aqua.ts
```diff
@@ -139,46 +142,48 @@ export const compileToFiles = async ({
...
+        if (targetType === "js") {
+          const aquaToJsPromise = aquaToJs(compilationResult, "js");
+
+          const sourcesPromise = (async () => {
+            return (await aquaToJsPromise)?.sources;
+          })();
+
+          const typesPromise = (async () => {
+            return (await aquaToJsPromise)?.types;
+          })();
```
akim-bow, line 177:
```suggestion
        if (targetType === "js") {
          const output = await aquaToJs(compilationResult, "js");

          const [sources, types] = [output?.sources, output?.types];
```

yarn.lock
akim-bow, line 1: No dependencies changed but yarn.lock is updated for some reason which could potentially lead to some dependency problem. Just a suggestion but it's better to update this file only in dedicated PRs.

src/lib/aqua.ts
shamsartem: For some reason I thought this would be impossible, but looks like flatMap is smart enough when working with async functions as callbacks

yarn.lock
shamsartem: I just do "yarn" when adding new dependency. Here "@fluencelabs/aqua-to-js" was added as new dependency. I think it's correct if yarn.lock is updated after new dependency is added

akim-bow: Oh, didn't notice package.json changes. Sorry.

akim-bow (approved):

PR #522, opened 2023-10-05
chore: Add install script and test

nahsi:


.github/workflows/install-test.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Checkout fluence-cli
+        uses: actions/checkout@v3
+        with:
+          repository: fluencelabs/cli
+
+      - name: Run install.sh script
+        run: ./install.sh
+
+      - name: Check fluence
+        run: ${FLUENCE_USER_DIR}/cli/bin/fluence --version
```
folex, line 41: I'd also check that `--help` contains all the commands. Because it often happens that one installs `fcli`, and it has `--version`, but has no other commands.

install.sh
```diff
@@ -0,0 +1,93 @@
+#! /usr/bin/env bash
```
folex, line 1:
```suggestion
#!/usr/bin/env bash
set -o errexit -o nounset -o pipefail

# set current working directory to script directory to run script from everywhere
cd "$(dirname "$0")"
```

.github/workflows/install-test.yml
nahsi: I'm checking there that installation script works, not fluence cli :)

folex: but it doesn't work if it installs fcli incorrectly, right

folex (approved):

PR #570, opened 2023-10-23
fix: env must not be required when using `--relay` [fixes DXJ-511]

shamsartem:


src/lib/jsClient.ts
```diff
@@ -29,25 +29,21 @@ import { resolveRelay } from "./multiaddres.js";
...
   const relay = await resolveRelay({
-    fluenceEnv,
     maybeFluenceConfig,
-    maybeRelay,
+    relayFromFlags,
+    fluenceEnvFromFlags,
   });
 
-  commandObj.logToStderr(
-    `Connecting to ${color.yellow(fluenceEnv)} relay: ${relay}`,
```
InversionSpaces, line 48: I believe this log is helpful to know what is happening. Could we leave it here and not print env if it is unknown or something like that?

InversionSpaces (approved):
utACK

PR #589, opened 2023-10-25
feat: allow user to select service from list [fixes DXJ-501]

shamsartem:


src/commands/service/repl.ts
```diff
@@ -248,4 +243,45 @@ const ensureModuleConfigsForToml = (
...
+      )} service from ${color.yellow(maybeFluenceConfig?.$getPath())}`,
+    }))
+  ) {
+    return firstService;
+  }
+
+  if (
+    serviceNames.length > 1 &&
+    (await confirm({
+      message: `Do you want to select a service from ${color.yellow(
```
folex, line 267: I don't see that user has a choice here, really. I think it makes sense to just show list of services, without asking to confirm fluence.yaml path.

shamsartem: I am not asking to confirm yaml path here, I am asking if he wants the service from the yaml or maybe he wants to enter the path himself or URL

Not completely sure it is useful but it's concivable that user might wan't to run some services not from fluence.yaml - I am just giving the way to do that

folex: But the alternative to "yes" is not clear. It's not obvious that "no" means "i want to specify url". I think it should be a single list with multiple options, including service names and url.

folex: That would save one y/n question, and make it easier to use overall :)

shamsartem: done

src/commands/service/repl.ts
```diff
@@ -243,45 +243,31 @@ const ensureModuleConfigsForToml = (
...
-      oneChoiceMessage() {
-        throw new Error("Unreachable");
-      },
-      onNoChoices() {
-        throw new Error("Unreachable");
-      },
+  const selected = await list({
+    message: "Select service",
+    options:
+      serviceNames.length === 0 ? [] : [...serviceNames, ENTER_PATH_OR_URL_MSG],
```
folex, line 257: Maybe suggest `ENTER_PATH_OR_URL_MSG` even if there are no services? Didn't get why not

shamsartem: I do it down below. If you leave it here it will ask you to select this one weird item in the list and only then you will have to actually enter the path or url

folex: got it, thx!

folex (approved):

PR #593, opened 2023-10-26
feat: Deal lifecycle

elshan-eth:
Stage configs:
Matcher contract: 0xCaDABD7dc67CDE6cA4B8Ff833F645105a6f86b1B

Local configs:
Matcher contract: 0x0e1F3B362E22B2Dc82C9E35d6e62998C7E8e2349


src/commands/deal/deposit.ts
```diff
@@ -0,0 +1,101 @@
...
+    const privKey = flags["priv-key"];
+
+    const dealAddress =
+      args["DEAL-ADDRESS"] ?? (await input({ message: "Enter deal address" }));
+
+    const amount = ethers.parseEther(
+      String(
+        args["AMOUNT"] ??
+          (await input({ message: "Enter amount of tokens to deposit" })),
+      ),
```
akim-bow, line 67: Why do you need casting to string? Isn't `input` function returns string already?

src/commands/provider/remove-units.ts
```diff
@@ -0,0 +1,96 @@
...
+      flags.units ?? (await input({ message: "Enter new units count" }));
+
+    const peerId =
+      flags["peer-id"] ?? (await input({ message: "Enter peerId" }));
+
+    const [{ DealClient }, { digest }, { base58btc }] = await Promise.all([
+      import("@fluencelabs/deal-aurora"),
+      import("multiformats"),
+      // eslint-disable-next-line import/extensions
+      import("multiformats/bases/base58"),
```
akim-bow, line 68: Maybe we can import multiformats statically?
@shamsartem

src/commands/provider/info.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    commandObj.log(color.gray(`--Peers--`));
+
+    const peerIdsAndPeers =
+      await matcher.getPeersByComputeProvider(providerAddress);
+
+    if (peerIdsAndPeers[0].length === 0) {
+      commandObj.log(color.gray(`No peers`));
+    } else {
+      for (let i = 0; i < peerIdsAndPeers[0].length; i++) {
```
akim-bow, line 93: Suggest to change this cycle to for of cycle for greater readability.
You can ground both array in pairs through `zip`
`for (const [p0, p1] =  lodash.zip(peerIdsAndPeers[0], peerIdsAndPeers[1])) {`

src/commands/provider/remove-units.ts
shamsartem: we can import everything statically, but the less static imports - the faster cli works as per my experiments

shamsartem: There is still stuff that is imported statically but I tried to remove what I can - and each lib that I removed improved speed a bit

shamsartem (approved):

PR #600, opened 2023-11-01
feat: migrate "workers" property in fluence.yaml directly into "deals" and "hosts" properties [fixes DXJ-507]

shamsartem:


docs/configs/fluence.md
```diff
@@ -16,7 +16,7 @@ Defines Fluence Project, most importantly - what exactly you want to deploy and
...
-| `hosts`                | [object](#hosts)            | No       | A map of objects with worker names as keys, each object defines a list of peer IDs to host the worker on                                                                                                                                                                                                                                                                                                                             |
+| `hosts`                | [object](#hosts)            | No       | A map of objects with worker names as keys, each object defines a list of peer IDs to host the worker on. If you want to deploy your services to specific peerIds. Intended to be used by providers to deploy directly without using the blockchain                                                                                                                                                                                  |
```
akim-bow, line 19:
```suggestion
| `hosts`                | [object](#hosts)            | No       | A map of objects with worker names as keys, each object defines a list of peer IDs to host the worker on. Intended to be used by providers to deploy directly without using the blockchain                                                                                                                                                                                  |
```
Probably redundant sentence. You already said that "list of peer IDs to host the worker on".

PR #603, opened 2023-11-02
chore: Check that wget is installed

nahsi:


install.sh
```diff
@@ -1,37 +1,68 @@
...
+  cat <<ERR
+fluence cli seems to be already installed.
+If it was installed with with npm please unistall it first with 'npm uninstall -g @fluencelabs/cli' then rerun installation script.
+If it was installed with installation script you should use 'fluence update' command to update to latest version.
+ERR
+  exit 1
+fi
+
 # Verify if the cli directory is empty
 if $(ls -1 ${FLUENCE_USER_DIR}/cli &>/dev/null); then
```
vasd85, line 39: This statement is TRUE when `${FLUENCE_USER_DIR}/cli` exists and empty.  Generally this statement will be false only when `ls -1` returns error. Maybe would be better to check for files in `${FLUENCE_USER_DIR}/cli` in some direct way?

vasd85: image

install.sh
```diff
@@ -41,14 +72,14 @@ url="https://fcli-binaries.s3.eu-west-1.amazonaws.com/channels/stable/${archive}
...
 echo "Fluence CLI installation complete!"
-if echo $PATH | grep -q $HOME/.local/bin ; then
+if echo $PATH | grep -q $HOME/.local/bin; then
   mkdir -p $HOME/.local/bin
   ln -sf ${FLUENCE_USER_DIR}/cli/bin/fluence $HOME/.local/bin/fluence
 else
-  echo "Create a symlink to fluence binary in any directory that is in \$PATH, for example:"
-  echo "sudo ln -s ${FLUENCE_USER_DIR}/cli/bin/fluence /usr/local/bin/fluence"
+  echo "Adding fluence cli to \$PATH with 'sudo ln -s ${FLUENCE_USER_DIR}/cli/bin/fluence /usr/local/bin/fluence"
+  sudo ln -s ${FLUENCE_USER_DIR}/cli/bin/fluence /usr/local/bin/fluence
```
vasd85, line 84: image
Maybe call to `ln -sf` here also, so as not to exit with the error?

vasd85 (changes requested):

install.sh
```diff
@@ -1,37 +1,68 @@
...
+if ! has "curl"; then
+  echo "'curl' is required for the installation. Please install it and rerun the script"
+  exit 1
+fi
+
+# Check that fluence is not installed
+if has "fluence"; then
+  cat <<ERR
+fluence cli seems to be already installed.
+If it was installed with with npm please unistall it first with 'npm uninstall -g @fluencelabs/cli' then rerun installation script.
```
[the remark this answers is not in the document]
nahsi: installed with npm
uninstall with ...

But wording is not perfect I agree. I will try to feed it to chatgpt :)

install.sh
nahsi: Nice catch! But I will change wording about directory being empty or not. If this directory exits we can assume that it was created by installation script only.

I can also do something like this
```
if [[ -d "${FLUENCE_USER_DIR}/cli" ]] && [[ -n "$(ls -A ${FLUENCE_USER_DIR}/cli)" ]]; then
    echo "Directory exists and is not empty"
else
    echo "Directory does not exist or is empty"
fi
```

But I think adding more code in that situation is not worth it.

install.sh
```diff
@@ -1,54 +1,88 @@
...
+if ! has "curl"; then
+  echo "Error: Installation requires 'curl'. Please install 'curl' and try again."
+  exit 1
+fi
+
+# Check that fluence is not installed
+if has "fluence"; then
+  cat <<ERR
+Error: Fluence CLI appears to be already installed.
+If it was installed with with npm please unistall it first with 'npm uninstall -g @fluencelabs/cli' and then rerun installation script.
```
nahsi, line 33:
```suggestion
If it was installed with npm please unistall it first with 'npm uninstall -g @fluencelabs/cli' and then rerun installation script.
```

install.sh
```diff
@@ -1,54 +1,88 @@
...
-if echo $PATH | grep -q $HOME/.local/bin ; then
+# Add Fluence CLI to path with symling
+if echo $PATH | grep -q $HOME/.local/bin; then
+  echo "Adding Fluence CLI symlink to ${HOME}/.local/fluence"
   mkdir -p $HOME/.local/bin
   ln -sf ${FLUENCE_USER_DIR}/cli/bin/fluence $HOME/.local/bin/fluence
 else
-  echo "Create a symlink to fluence binary in any directory that is in \$PATH, for example:"
-  echo "sudo ln -s ${FLUENCE_USER_DIR}/cli/bin/fluence /usr/local/bin/fluence"
+  echo "Adding Fluence CLI symling to /usr/local/bin/fluence"
```
nahsi, line 85:
```suggestion
  echo "Adding Fluence CLI symlink to /usr/local/bin/fluence"
```

vasd85 (approved):

PR #604, opened 2023-11-03
feat: restructure project templates [fixes DXJ-523]

shamsartem:


test/setupTests.ts
```diff
@@ -48,23 +52,21 @@ try {
   });
 } catch {}
 
-const [firstPath, ...rest] = await Promise.all(
+const [, ...rest] = await Promise.all(
```
akim-bow, line 55:
```suggestion
const templates = await Promise.all(
```

shamsartem: but that's completely different. in my case I use everything except the first one

akim-bow: sry, didn't notice comma
i meant `const [, templates] = await Promise.all(`
because `rest` variable name is not very self-descriptive

akim-bow (approved):

PR #608, opened 2023-11-08
feat: restructure templates [fixes DXJ-526]

shamsartem:


src/lib/init.ts
```diff
@@ -312,92 +295,228 @@ type InitTSorJSProjectArg = {
...
+        (() => {
+          throw new Error(\`Button with id="\${fnName}" is not found\`);
+        })();
+
+      buttonEl.addEventListener("click", () => {
+        runAquaFunction(fn);
+      });
+    });
+  });
+})();
```
akim-bow, line 430: No need to use `requestAnimationFrame`  optimization here i think. You make very simple operation and just once. That function usually used for DOM animation which optimized to be updated only when your monitor showing next frame.

src/lib/init.ts
```diff
@@ -312,92 +295,228 @@ type InitTSorJSProjectArg = {
...
+    })(),
+  ]);
+}
+
+function getIndexHTMLContent(isJS: boolean) {
+  return `<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" href="data:,">
```
akim-bow, line 346: didn't understand this. Maybe we can add fluence icon or something here?

shamsartem: we would have to add icon code to template as well, we probably can do it. Chrome always requests favicon.ico even if you remove this and you get 404, this is the simplest way to make chrome shut up

shamsartem: inserted base64 encoded image, seems fine

src/lib/init.ts
shamsartem: restructured this part with createElement like in other places in the code so there is no need to find by id and stuff

PR #610, opened 2023-11-09
feat!: use Config tomls for nox configuration [fixes DXJ-519]

shamsartem:


docs/configs/provider.md
```diff
@@ -22,11 +23,102 @@ Defines config used for provider set up
...
+| `websocketPort`  | number                    | No       | Both host and container WebSocket port to use. Default: for each nox a unique port is assigned starting from 9991 |
+
+### systemServices
+
+#### Properties
+
+| Property   | Type                | Required | Description                       |
+|------------|---------------------|----------|-----------------------------------|
+| `aquaIpfs` | [object](#aquaipfs) | No       |                                   |
+| `decider`  | [object](#decider)  | No       |                                   |
```
akim-bow, line 97: ditto

src/lib/configs/initConfig.ts
```diff
@@ -547,17 +547,7 @@ export function getConfigInitFunction<
...
-            `Couldn't save config ${color.yellow(
-              configPath,
-            )}. ${await validationErrorToString(
-              initializedReadonlyConfig.$validateLatest.errors,
-            )}`,
-          );
-        }
-
-        const config = removeProperties({ ...this }, ([, v]) => {
+        const config = removeProperties({ ...(await this) }, ([, v]) => {
```
akim-bow, line 550: Please explain this line for me.
If i understand this code correctly, `this` points out to the returned object.
Why we need `await` then?
Also it would be good to add a `type` for this here

src/lib/configs/project/dockerCompose.ts
```diff
@@ -269,11 +329,11 @@ async function genDockerCompose(
...
     },
-    secrets: Object.fromEntries(
-      peers.map(({ name, path: file }) => {
-        return [name, { file }];
+    secrets: Object.fromEntries([
+      ...peers.map(({ name, relativeSecretFilePath: file }) => {
+        return [name, { file }] as const;
       }),
-    ),
+    ]),
```
akim-bow, line 336: Is there are some ts errors that made you to use spread operator here?
Probably should work without spread.

src/lib/configs/project/dockerCompose.ts
```diff
@@ -329,3 +401,162 @@ export const initReadonlyDockerComposeConfig =
...
+      return writeFile(
+        join(configsDir, getConfigTomlName(key)),
+        [
+          stringify(configYAMLToConfigToml(overridden)),
+          providerConfig.nox?.rawConfig ?? "",
+          value.nox?.rawConfig ?? "",
+        ]
+          .filter((x) => {
+            return x !== "";
+          })
```
akim-bow, line 442:
```suggestion
          .filter(Boolean)
```

akim-bow: as far as i remember this is one of the case in `total-typescript`

src/lib/configs/project/dockerCompose.ts
```diff
@@ -329,3 +401,162 @@ export const initReadonlyDockerComposeConfig =
...
+            },
+            decider: {
+              ...a.systemServices.decider,
+              ...b.systemServices.decider,
+            },
+          },
+        }
+      : {}),
+  };
+}
```
akim-bow, line 477: OMG! Can we use something like `lodash.merge` ?

src/lib/configs/project/dockerCompose.ts
```diff
@@ -329,3 +401,162 @@ export const initReadonlyDockerComposeConfig =
...
+
+function configYAMLToConfigToml(config: NoxConfigYAML) {
+  const configToml: NoxConfigTOML = {};
+
+  if (config.aquavmPoolSize !== undefined) {
+    configToml.aquavm_pool_size = config.aquavmPoolSize;
```
akim-bow, line 483: OMG 2 
Can we use something like `https://www.npmjs.com/package/change-case`

src/lib/configs/project/fluence.ts
```diff
@@ -159,6 +160,7 @@ const serviceSchema: JSONSchemaType<ServiceV1> = {
     },
   },
   required: ["get"],
+  additionalProperties: false,
```
akim-bow, line 163: Is there are lib to generate JSONSchema from decorator annotations on an runtime object itself?
Would benefit bcs that would keep object strucure in a single place

src/versions.json
```diff
@@ -1,5 +1,5 @@
 {
-  "nox": "fluencelabs/nox:0.4.0",
+  "nox": "docker.fluence.dev/nox:FLU-494",
```
akim-bow, line 2: Remember to revert it after merge

.github/workflows/tests.yml
```diff
@@ -109,8 +109,8 @@ jobs:
           # if nox image is not passed from e2e read from versions.json
           'null')
             nox="$(jq .nox src/versions.json -r)"
-            # use minimal image
-            nox="${nox//nox:/nox:minimal_}"
+            # # use minimal image
```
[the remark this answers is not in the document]
shamsartem: this will be all removed when Anatoly makes the release

docs/configs/provider.md
```diff
@@ -22,11 +23,102 @@ Defines config used for provider set up
...
 
-| Property        | Type   | Required | Description                                                                                                       |
-|-----------------|--------|----------|-------------------------------------------------------------------------------------------------------------------|
-| `computeUnits`  | number | No       |                                                                                                                   |
-| `tcpPort`       | number | No       | Both host and container TCP port to use. Default: for each nox a unique port is assigned starting from 7771       |
-| `webSocketPort` | number | No       | Both host and container WebSocket port to use. Default: for each nox a unique port is assigned starting from 9991 |
+| Property       | Type           | Required | Description |
+|----------------|----------------|----------|-------------|
+| `computeUnits` | number         | No       |             |
+| `nox`          | [object](#nox) | No       |             |
```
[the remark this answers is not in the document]
shamsartem: Yeah missed the object themselves, I will put the descriptions

src/lib/configs/initConfig.ts
shamsartem: I mostly put it here to shut up typescript. It thinks this can be a promise because the function around it might return a promise. Doesn't really matter much if there is await here or not

src/lib/configs/project/dockerCompose.ts
shamsartem: I first used merge from lodash and find out it doesn't merge the way I want it to, but I checked out and there is a mergeWith which I can use so Arrays are overrided instead of weirdly merged (by default in arrays items are overriden by index so if you have `prop:  ['a', 'b']` in one config and `prop: ['c']` in another you will get `prop: ['c', 'b']`. I think it's better to just get `prop: ['c']` in the end result because we want users to be able to override configs completely)

src/lib/configs/project/dockerCompose.ts
shamsartem: there is the function to change case in lodash but I didn't use it because 
first it wasn't too big when I started
second I wanted to retain some type information
third I'v been thinking that not everything will be as straight forward as `camelCase` -> `snake_case`

but seems it got a bit out of hand, you are right, It will be less error-prone if we don't care about types and just transform this to snake_case. If there will be some differenences in the future - then I think I will extend this function so it does some custom transaformations when and if they will be required in the future

src/versions.json
shamsartem: I will revert this before merge

src/lib/configs/initConfig.ts
folex: wow, looks fun!

can you put that to comment, plz?

src/lib/configs/project/dockerCompose.ts
```diff
@@ -45,11 +59,68 @@ import {
...
   initNewReadonlyProviderConfig,
   type ProviderConfigReadonly,
 } from "./provider.js";
 
+const NOX_IPFS_MULTIADDR = `/dns4/${IPFS_CONTAINER_NAME}/tcp/${IPFS_PORT}`;
+
+function getIsLocal() {
+  if (envConfig?.fluenceEnv === undefined) {
+    return true;
+  }
```
folex, line 75: Why do you interpret empty `env.yml` to mean `env: local`?

src/lib/configs/project/dockerCompose.ts
```diff
@@ -45,11 +59,68 @@ import {
...
+  const isLocal = getIsLocal();
+
+  const contractsEnv: ContractsENV =
+    envConfig?.fluenceEnv === "custom"
+      ? fluenceConfig?.customFluenceEnv?.contractsEnv ?? "local"
+      : envConfig?.fluenceEnv ?? "local";
+
+  const { DEAL_CONFIG } = await import(
+    "@fluencelabs/deal-aurora/dist/client/config.js"
+  );
```
folex, line 92: nice

src/lib/configs/project/dockerCompose.ts
```diff
@@ -45,11 +59,68 @@ import {
...
+        workerIpfsMultiaddr: isLocal
+          ? NOX_IPFS_MULTIADDR
+          : "http://ipfs.fluence.dev",
+        networkApiEndpoint: isLocal
+          ? `http://${CHAIN_CONTAINER_NAME}:${CHAIN_PORT}`
+          : "http://mumbai-polygon.ru:8545",
+        networkId: dealConfig.chainId,
+        startBlock: "earliest",
+        // TODO: use correct addr for env
+        matcherAddress: "0x0f68c702dC151D07038fA40ab3Ed1f9b8BAC2981",
```
folex, line 118: Gotta take it from `deal` package

src/lib/configs/project/dockerCompose.ts
```diff
@@ -329,3 +403,96 @@ export const initReadonlyDockerComposeConfig =
...
+      }
+
+      if (overridden.httpPort === undefined) {
+        overridden.httpPort = HTTP_PORT_START + i;
+      }
+
+      if (
+        "systemServices" in overridden &&
+        "decider" in overridden.systemServices &&
+        overridden.systemServices.decider.walletKey === undefined
```
folex, line 438: I don't understand the logic here.

Why not like this?
```
if overridden?.systemServices?.decider?.walletKey === undefined
```

In this case, it would write wallet key even if there are no `systemServices` or `systemServices.decider` in the `overridden`. But right now it doesn't, and I'm not sure why.

src/lib/configs/project/provider.ts
```diff
@@ -46,20 +49,165 @@ export type Offer = {
...
+              type: "string",
+              description: `Start block`,
+            },
+            matcherAddress: {
+              nullable: true,
+              type: "string",
+              description: `Matcher address`,
+            },
+            walletKey: {
+              nullable: true,
```
folex, line 174: I think it should be `nullable: false`, because Decider doesn't have a default value for `walletKey`

src/lib/const.ts
```diff
@@ -119,6 +119,8 @@ export const CHAIN_CONTAINER_NAME = "chain";
 export const CHAIN_PORT = 8545;
 export const TCP_PORT_START = 7771;
 export const WEB_SOCKET_PORT_START = 9991;
+export const HTTP_PORT_START = 18080;
+export const DEFAULT_AQUAVM_POOL_SIZE = 8;
```
folex, line 123: That's a bit too high for one of three docker containers ran on a local machine, I think. `2` should do.

folex (approved):

src/lib/configs/project/dockerCompose.ts
shamsartem: That's a good catch, I actually don't remember why I did this to be honest. Probly wanted to try something out quickly and forgot to fix it. Anyway I am thinking that with multiple porvider.yaml in one dir we would have to set `env` right in provider.yaml itself, though I previously thought this would be unnecessary. So will remove this part and get env from provider config

src/lib/configs/project/dockerCompose.ts
shamsartem: I'v been thinking that there is probably a chance that user will not want to have decider at all for the local test.
If we do it like you are suggesting - then this property will be set in all cases. I don't know, Maybe we can do it like you suggest, cause we probably better of not thinking about the cases when there is no decider. What do you think?

src/lib/configs/project/provider.ts
shamsartem: for now if user omits it - cli will pick one of the keys from the local network, but we should setup some way of generating these keys. We discussed using files previously. Maybe files should be used and passed to containers and this property should be removed completely from here, I just temporarely did this so tests work at least

src/lib/const.ts
shamsartem: Good. Will change

src/lib/configs/project/dockerCompose.ts
folex: Hm, but `provider.yml` won't set Env for the Developer, right?

ie, imagine a project with _just_ `fluence.yml`, where would you store the selected env?

or maybe I understood your thinking wrong :)

src/lib/configs/project/dockerCompose.ts
folex: Yeah, I think that Decider should be enabled by default. Otherwise `deal deploy` won't work.

src/lib/configs/project/fluence.ts
shamsartem: not sure about that. I would only say that it's most flexible to write the schema by hand using the schema syntax itself without any additional wrappers/tools. This way a lot of things can be controled and some of the things can be shared easilly between config versions. Also typesacript helps to write this schemas a bit. It is not the best experience to use this schema lib for me and to maintain all that, I will tell you that. But I find it's worth it because users can enjoy good autocomplete validation and docs for configs right inside their IDE

src/lib/configs/project/provider.ts
folex: I think that's OK.

> cli will pick one of the keys from the local network

I understand this as fcli has pre-generated list of keys somewhere. If my understanding is correct, what will happen if a user generates more Noxes than you have pre-generated keys?

src/lib/configs/project/dockerCompose.ts
shamsartem: provider.yaml will not set env for developer. It will only set env for itself so when provider generates stuff from it - it's generated for the correct env.

developer will still use env.yaml under the hood as before when working with the project. and for local development `env: local` will be preset in provider.yaml

src/lib/configs/project/provider.ts
shamsartem: currently it will cycle through them and use the same for multipple peers, I am thinking to maybe add a warning about that when I put the keys in the files. But probably when running local env users will rarely need more than 10 and we have 20 hardcoded keys

folex: Maybe it's easier to generate keys? I think you can simply use 32-byte array as a Private Key, and increment last numbers in it.

src/lib/configs/initConfig.ts
akim-bow: I suggest just type this value manually instead of putting comments or something
async $commit(): Promise<void> {
->
`async $commit(this: InitializedConfig<LatestConfig> | null): Promise<void> {`
It will make sure that `this` will never be `Promise` value and ts will tell you that if not.

akim-bow: Eventually i made something like this
```
async $commit(this: InitializedConfig<LatestConfig>): Promise<void> {
        const config = removeProperties(this, ([, v]) => {
          return typeof v === "function";
        });
```

src/lib/configs/project/provider.ts
shamsartem: I can ofcourse generate keys, but will it be a problem in the furture that these keys don't have any money on their account? The ones that I use have some by default
If it's always ok for compute peers to have empty accounts - then generating seems to be a better option

src/lib/configs/initConfig.ts
shamsartem: I used your suggestion since it doesn't leave any useless code in runtime. What you suggest is essentially typecasting (but for some reason eslint doesn't consider this to be a typescasting even though it clearly is). Anyway I left a comment still because this typecasting requires at least some explanation. And you correctly noticed that there is no need for spread operator here since `removeProperties` creates a new object anyway

akim-bow (approved):

PR #630, opened 2023-11-17
feat: Use js-client 0.5.0

akim-bow:


src/commands/run.ts
```diff
@@ -18,7 +18,7 @@ import { access, readFile } from "node:fs/promises";
 import { dirname, resolve } from "node:path";
 // import { performance, PerformanceObserver } from "node:perf_hooks";
 
-import type { callAquaFunction } from "@fluencelabs/js-client";
+import { v5_callFunction } from "@fluencelabs/js-client";
```
shamsartem, line 21:
```suggestion
import type { v5_callFunction } from "@fluencelabs/js-client";
```

src/commands/run.ts
```diff
@@ -405,16 +405,15 @@ const fluenceRun = async (args: RunArgs) => {
...
-    args: args.runData ?? {},
-    config: {},
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-    def: functionCall.funcDef,
-    peer: Fluence.getClient(),
-    script: functionCall.script,
-  });
+  const { Fluence, v5_callFunction } = await import("@fluencelabs/js-client");
+
+  const runArgs = Object.values(args.runData ?? []);
```
shamsartem, line 410: These are named arguments and here you rely on order of them instead of relying on their names as it worked previously. This will break if people don't order their arguments correctly

.yarnrc.yml
```diff
@@ -5,3 +5,4 @@ plugins:
     spec: "@yarnpkg/plugin-workspace-tools"
 
 yarnPath: .yarn/releases/yarn-3.6.4.cjs
+npmRegistryServer: https://npm.fluence.dev/
```
shamsartem, line 8: maybe this one now also needs to be removed?

src/commands/run.ts
```diff
@@ -405,18 +406,56 @@ const fluenceRun = async (args: RunArgs) => {
...
+
+  const schema = functionCall.funcDef;
+
+  // TODO: remove this after DXJ-535 is done
+  const validatedRunData = Object.fromEntries(
+    Object.entries(args.runData ?? {}).map(([argName, argValue]) => {
+      assert(
+        typeof argValue !== "function",
+        "Should be impossible to pass function as an argument to fluence run",
+      );
```
shamsartem, line 422: instead of this better change the RunData type to `Record<string, Parameters<typeof js2aqua>[0]>`

src/commands/run.ts
```diff
@@ -405,18 +406,56 @@ const fluenceRun = async (args: RunArgs) => {
...
+
+      const fields =
+        schema.arrow.domain.tag === "nil" ? {} : schema.arrow.domain.fields;
+
+      const argSchema = fields[argName];
+
+      assert(
+        argSchema !== undefined,
+        "Should be impossible because schema always contains same keys as runData",
+      );
```
shamsartem, line 432: but if schema.arrow.domain.tag === "nil" then fields would be empty object and this would be undefined
when schema.arrow.domain.tag === "nil" is nil?

src/commands/run.ts
```diff
@@ -405,18 +406,56 @@ const fluenceRun = async (args: RunArgs) => {
...
   });
 
-  return result;
+  const returnSchema =
+    schema.arrow.codomain.tag === "unlabeledProduct" &&
+    schema.arrow.codomain.items.length === 1
+      ? schema.arrow.codomain.items[0]
+      : schema.arrow.codomain;
+
+  assert(returnSchema !== undefined, "This value cannot be 'undefined'");
```
shamsartem, line 456: if it's impossible, better explain why

src/commands/run.ts
```diff
@@ -405,18 +406,56 @@ const fluenceRun = async (args: RunArgs) => {
...
-    script: functionCall.script,
+    args: validatedRunData,
   });
 
-  return result;
+  const returnSchema =
+    schema.arrow.codomain.tag === "unlabeledProduct" &&
+    schema.arrow.codomain.items.length === 1
+      ? schema.arrow.codomain.items[0]
+      : schema.arrow.codomain;
```
shamsartem, line 454: can be done like this so there is no need for asserts
```suggestion
  const returnSchema =
    (schema.arrow.codomain.tag === "unlabeledProduct"
      ? schema.arrow.codomain.items[0]
      : undefined) ?? schema.arrow.codomain
```

akim-bow: Won't work. It's different code. What if length property is `2`?

src/commands/run.ts
akim-bow: If `schema.domain.tag` would equal "nil", then this `map` functions wouldn't even execute a single time.

src/commands/run.ts
akim-bow: undefined appears in union after making checked indexed access to an array `items`.
Length of this has been checked already.
To me it looks ridiculous to put an assert after every array or object access.
Useless runtime work.

akim-bow: Updated description there

src/commands/run.ts
shamsartem: yeah didn't correctly understand the code

shamsartem: That's kinda of a problem that I don't really understand what is happening here, why you here care only about the case when there is 1 item

src/commands/run.ts
shamsartem: map would execute if person puts anything in runData, user can accidentally do that easilly

src/commands/run.ts
```diff
@@ -405,18 +406,54 @@ const fluenceRun = async (args: RunArgs) => {
...
+  const validatedRunData = Object.fromEntries(
+    Object.entries(args.runData ?? {}).map(([argName, argValue]) => {
+      const fields =
+        schema.arrow.domain.tag === "nil" ? {} : schema.arrow.domain.fields;
+
+      const argSchema = fields[argName];
+
+      assert(
+        argSchema !== undefined,
+        "Should be impossible because schema always contains same keys as runData",
```
shamsartem, line 426: It's totally possible. User can put anything in runData even accidentally use incorrect name. We should validate this and give a good error message if user does something wrong

src/commands/run.ts
```diff
@@ -405,18 +406,54 @@ const fluenceRun = async (args: RunArgs) => {
...
+      );
+
+      assert(
+        argSchema.tag !== "arrow",
+        "Should be impossible to pass function as an argument to fluence run",
+      );
```
shamsartem, line 432: If user tries to run the function that accepts arrow as an argument using `fluence run` we should give a normal error message for hime that explains that it's not possible to do

src/commands/run.ts
shamsartem:
```suggestion
const returnSchema =
    (schema.arrow.codomain.tag === "unlabeledProduct" &&
    schema.arrow.codomain.items.length === 1
      ? schema.arrow.codomain.items[0]
      : undefined) ?? schema.arrow.codomain;
```

shamsartem (approved):

PR #635, opened 2023-11-23
feat: move some flags into configs for provider and deploy

shamsartem:


docs/commands/README.md
```diff
@@ -231,19 +231,17 @@ Deploy workers according to deal in 'deals' property in fluence.yaml
 
 ```
 USAGE
-  $ fluence deal deploy [WORKER-NAMES] --collateral-per-worker <value> --min-workers <value> --target-workers
-    <value> --max-workers-per-provider <value> --price-per-worker-epoch <value> [--no-input] [-k <value>]
-    [--off-aqua-logs] [--priv-key <value>] [--env <value>] [--relay <value>] [--ttl <value>] [--dial-timeout <value>]
-    [--particle-id] [--import <value>] [--no-build] [--tracing] [--marine-build-args <value>] [--auto-match]
```
akim-bow, line 237: kaef

src/commands/deal/create.ts
```diff
@@ -66,11 +66,11 @@ export default class Create extends BaseCommand<typeof Create> {
 
     const dealAddress = await dealCreate({
       appCID: flags["app-cid"],
-      collateralPerWorker: flags["collateral-per-worker"],
+      collateralPerWorker: Number(flags["collateral-per-worker"]),
       minWorkers: flags["min-workers"],
       targetWorkers: flags["target-workers"],
       maxWorkersPerProvider: flags["max-workers-per-provider"],
-      pricePerWorkerEpoch: flags["price-per-worker-epoch"],
+      pricePerWorkerEpoch: Number(flags["price-per-worker-epoch"]),
```
akim-bow, line 73: Do you validate it somehow? What if string contains float value for example?

src/commands/deal/deploy.ts
```diff
@@ -162,8 +142,19 @@ export default class Deploy extends BaseCommand<typeof Deploy> {
       definition: appCID,
     } of uploadResult.workers) {
       const deal = fluenceConfig.deals?.[workerName];
-      assert(deal !== undefined);
-      const { minWorkers = MIN_WORKERS, targetWorkers = TARGET_WORKERS } = deal;
+
+      assert(
+        deal !== undefined,
+        "Unreachable. All worker names are checked above",
+      );
```
akim-bow, line 149: I don't like this assert. If you made validation already, probably your validating function should return refined `fluenceConfig.deals` like a proof of validation? Probably every validation should either throw or return refined value e. g. without union with `undefined` or `null`.

src/lib/configs/project/provider.ts
```diff
@@ -431,11 +449,19 @@ export async function initNewReadonlyProviderConfig({
...
+export function initProviderConfig(name?: string | undefined) {
+  if (name !== undefined) {
+    setProviderConfigName(name);
+  }
+
   return getConfigInitFunction(getInitConfigOptions())();
 }
 
-export function initReadonlyProviderConfig() {
+export function initReadonlyProviderConfig(name?: string | undefined) {
```
akim-bow, line 460:
```suggestion
export function initReadonlyProviderConfig(name?: string) {
```

Looks like in js omitting argument is the same as passing `undefined` directly

src/commands/deal/create.ts
shamsartem: this command is hidden from users and only for development purposes, so I don't validate anything here

src/lib/configs/project/provider.ts
shamsartem: Yeah for function params this seems to be true, will apply this imporvement. This is not true for objects though, cause for objects a new key might be created with undefined inside, which would a different runtime behaviour

akim-bow:
Looks nice!

PR #637, opened 2023-11-24
fix: improve errors for configs

shamsartem:


kmd-fl (approved):

src/lib/ajvInstance.ts
```diff
@@ -37,14 +42,34 @@ export const validationErrorToString = async (errors: AjvErrors) => {
...
     "Errors:\n\n" +
     errors
-      .filter(({ instancePath }) => {
-        return instancePath !== "";
-      })
-      .map(({ instancePath, params, message }) => {
+      .map(({ instancePath, params, message }, i) => {
         const paramsMessage = yamlDiffPatch("", {}, params);
-        return `${color.yellow(instancePath)} ${
+        const prev = errors[i - 1];
```
akim-bow, line 47: I think `prevError` would be better naming

src/lib/ajvInstance.ts
```diff
@@ -37,14 +42,34 @@ export const validationErrorToString = async (errors: AjvErrors) => {
...
+        const isDuplicateError =
+          prev?.instancePath === instancePath &&
+          jsonStringify(prev.params) === jsonStringify(params) &&
+          prev.message === message;
+
+        if (isPreviousVersionError || isDuplicateError) {
+          return "";
+        }
+
+        return `${instancePath === "" ? "" : `${color.yellow(instancePath)} `}${
```
akim-bow, line 64: Probably `intancePath || color.yellow...`would be simpler? Or you have other reason to use ternary here?

src/lib/ajvInstance.ts
```diff
@@ -37,14 +42,34 @@ export const validationErrorToString = async (errors: AjvErrors) => {
...
           message ?? ""
-        }\n${paramsMessage}`;
+        }${paramsMessage === "" ? "" : `\n${paramsMessage}`}`;
```
akim-bow, line 66: Same

src/lib/ajvInstance.ts
```diff
@@ -37,14 +42,34 @@ export const validationErrorToString = async (errors: AjvErrors) => {
...
+      })
+      .filter((s) => {
+        return (
+          s !== "" &&
```
akim-bow, line 70: `s && s !== "must match..."`

src/lib/ajvInstance.ts
shamsartem: Linter forbids using strings in the places where bool is expected to prevent unexpected behavior and to be more explicit about what exactly the check is checking

shamsartem: Because e.g. sometimes empty string is valid and you want to get rid of only undefined case

src/lib/ajvInstance.ts
```diff
@@ -37,14 +42,34 @@ export const validationErrorToString = async (errors: AjvErrors) => {
...
+          s !== "must match exactly one schema in oneOf\npassingSchemas: null\n"
+        );
```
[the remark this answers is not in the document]
shamsartem: If the error changes then user will start seeing a new useless error, I am removing it here only because it's useless. I don't think moving it to const will help, but probably writing a test could, but I think writing a test for this would be too much, this bit is not so important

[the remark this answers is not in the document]
shamsartem: Can move it to const just for little bit of documentation purpose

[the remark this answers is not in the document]
akim-bow: But can you control this error on ajv? E.g. intercept it before error reaches the user. For example in js-client i use zod which allows me to intercept errors by filtering their codes and classes, change their message and even silence them.
[Zod error API](https://zod.dev/ERROR_HANDLING).
Because if i understand the code correctly, you want to prevent certain kind of errors from getting in user console.

src/lib/configs/initConfig.ts
```diff
@@ -277,6 +287,9 @@ export function getReadonlyConfigInitFunction<
       getSchemaDirPath,
     } = options;
 
+    // every config schema has a version property by convention
+    // eslint-disable-next-line
+    const latestConfigVersion = latestSchema.properties.version.const as string;
```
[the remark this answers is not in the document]
shamsartem: I actually do extend schema, so it's impossible to pass pass a schema to this function that doesn't have version. But for some reason ajv types are not perfect in this sence and they don't produce correct results. I will improve comment to explains this a bit better. If you have nothing interesting to do you can to resolve as well, I wasnt' able to do it in a clean way

src/lib/ajvInstance.ts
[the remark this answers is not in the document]
akim-bow: Maybe you can give it a try soon or later with TODO and separate task. Dunno how hard would be to rewrite such places (if it even possible).

PR #640, opened 2023-11-29
feat: create a working local env automatically [fixes DXJ-528]

shamsartem:


src/lib/helpers/utils.ts
```diff
@@ -134,3 +134,40 @@ export function splitErrorsAndResults<T, U, V>(
 
   return [errors, results] as const;
 }
+
+export async function setTryTimeout<T, U>(
```
akim-bow, line 138: I suggest to use something like [this](https://www.npmjs.com/package/retry) instead

src/lib/configs/project/provider.ts
```diff
@@ -326,8 +330,30 @@ function getDefault(args: Omit<ProviderConfigArgs, "name">) {
...
+          .fill(null)
+          .map((_, i) => {
+            return [
+              `nox-${i}`,
+              {
+                computeUnits: 1,
+              },
+            ] as const;
+          }),
+      );
```
akim-bow, line 345: You have lodash, right?
You can use `count(25).` instead of new `Array(25).fill(null).`

shamsartem: I always forget I have lodash. I think it's called `times` though, not `count`

src/lib/helpers/utils.ts
akim-bow: Your solution has following drawback.

It does not support attempt limits additionally to just timeout. It is important in cases if function fails immediately. If attempts not fixated, your function could be called million of times, greatly reducing cli performance.
Additionally it's not optimized - timeout will be added **before** every call, but should be added **between** calls.

I think there are more cases if you compare your function to lib implementation.

Writing this very common function by yourself just pointless in my opinion. If you came to the need to use retry pattern, lets use it in a production ready and optimized way in a dedicated lib to save your precious time.

folex: If we can avoid introducing more deps without any meaningful drawbacks, that's great.

akim-bow: Your function allows to be called without `msBetweenTries` so it's possible to timeout. I would be ok it this is required argument then.

Also, i repeat, you introduced timeout before even the 1 call - 
```
if (msBetweenTries !== undefined) {
      await new Promise((resolve) => {
        setTimeout(resolve, msBetweenTries);
      });
    }
```
so it will add a delay even in succesful execution.

After i saw these details, i'm sure that you and i will save a lot of time if we just use 0 deps and very lightweight solutions instead of developing our own Fluence baked bicycle.

akim-bow: Also the code would be much more readable if you use instead of custom wrappers popular solutions covered be documentation.

akim-bow (approved):

PR #648, opened 2023-12-12
feat: use new aqua packages [fixes DXJ-534, DXJ-529, DXJ-478, DXJ-370]

shamsartem:


src/beforeBuild.ts
```diff
@@ -15,12 +15,28 @@
...
+  NODE_MODULES_DIR_NAME,
+} from "./lib/const.js";
+import { versions } from "./versions.js";
+
+const aquaDependenciesDirPath = join("src", AQUA_DEPENDENCIES_DIR_NAME);
+await mkdir(aquaDependenciesDirPath, { recursive: true });
+
+await writeFile(
+  join(aquaDependenciesDirPath, "package.json"),
+  JSON.stringify({ dependencies: versions.npm }, null, 2),
```
akim-bow, line 37: I thought you had a function for this `jsonStringify`

akim-bow: Or you don't want to import something from CLI in this helper script?

src/commands/dep/cargo/install.ts
```diff
@@ -68,37 +65,31 @@ export default class Install extends BaseCommand<typeof Install> {
   };
 
   async run(): Promise<void> {
-    const { args, flags, maybeFluenceConfig } = await initCli(
+    const { args, flags, fluenceConfig } = await initCli(
```
akim-bow, line 68: I like this naming much more :)

akim-bow: What about the renaming in other commands?

src/commands/dep/reset.ts
```diff
@@ -14,111 +14,34 @@
...
-    const { maybeFluenceConfig, flags } = await initCli(
-      this,
-      await this.parse(Reset),
-    );
-
-    if (flags.global) {
-      const removedDependencies = await removeRecommendedDependencies(
-        userConfig.dependencies,
-      );
+    const { maybeFluenceConfig } = await initCli(this, await this.parse(Reset));
```
akim-bow, line 31: In some files you're using `fluenceConfig`, in other `maybeFluenceConfig`

src/commands/run.ts
```diff
@@ -373,7 +367,7 @@ type RunArgs = FromFlagsDef<(typeof Run)["flags"]> & {
   maybeFluenceConfig: FluenceConfig | null;
   funcCall: string;
   filePath: string;
-  imports: string[];
+  import: undefined | string[];
```
akim-bow, line 370:
```suggestion
  import: string[] | undefined;
```

akim-bow: Negligible suggestion

src/lib/deployWorkers.ts
```diff
@@ -136,24 +137,28 @@ type UploadDeploySpellConfig =
 type PrepareForDeployArg = {
   workerNames: string | undefined;
   fluenceConfig: FluenceConfig;
-  aquaImports: Array<string>;
-  noBuild: boolean;
-  marineBuildArgs: undefined | string;
+  flags: {
+    import: Array<string> | undefined;
+    "no-build": boolean;
+    "marine-build-args": undefined | string;
```
akim-bow, line 143:
```suggestion
    "marine-build-args": string | undefined;
```

src/lib/npm.ts
```diff
@@ -108,125 +113,119 @@ export const getLatestVersionOfNPMDependency = async (
...
-};
+export async function npmUninstall({
+  packageNameAndVersion,
+  fluenceConfig,
+}: NpmUninstallArgs) {
+  assertDependenciesPresent(fluenceConfig);
+  const [packageName] = splitPackageNameAndVersion(packageNameAndVersion);
+
+  await runNpm({
+    args: ["un", packageNameAndVersion],
```
akim-bow, line 209:
```suggestion
    args: ["uninstall", packageNameAndVersion],
```
Looks a bit clearer.

test/integration.test.ts
```diff
@@ -42,24 +34,24 @@ describe("integration tests", () => {
...
+  //       i: "smoke.aqua",
+  //       quiet: true,
+  //     },
+  //     cwd,
+  //   });
+
+  //   const parsedResult = JSON.parse(result);
+  //   // Peer.identify() is supposed to return an object with air_version key
+  //   expect(parsedResult).toHaveProperty("air_version");
+  // });
```
akim-bow, line 56: Why did you comment this? Just curious.
Is the test does nothing in the new code and you want to persist it for some reason?

src/beforeBuild.ts
shamsartem: yeah don't want to import that, it's the reason

src/commands/dep/cargo/install.ts
shamsartem: it's currently if you pass "true" the command will make sure you have fluence project. If you don't pass that - it will possibly be null. So it's not a renaming, I just changed the requirements for this command, because now global dependency override is removed

test/integration.test.ts
shamsartem: The reason is that I have a bit of a problem
I want cli to be able to recognize dependencies in a convenient way.
Normally it's fluence.yaml where you put all the dependencies
But if you are creating a library, you probably want just to directly use npm, so you can publish, pack and do other npm stuff in CI etc.
So added some code that would find a nearest package.json and use that dir for resolving dependencies in case fluence.yaml is not present

This test was for when there is no project, but I currently always have package.json top level so I can't have a no-project situation anymore with this design

Maybe you have any ideas on how to continue with this, or we can probably just not have this test, assume users always have fluence project when they compile aqua. I don't know

shamsartem: restored this test by temporarelly removing package.json

akim-bow:
Nice work. Thanks!

akim-bow (approved):

PR #649, opened 2023-12-13
fix: raw config to be parsed, improve local up and service new [fixes DXJ-564]

shamsartem:


src/lib/configs/project/provider.ts
```diff
@@ -549,22 +549,30 @@ export async function ensureConfigToml(
...
   await Promise.all(
     computePeers.map(async ([computePeerName, computePeerConfig], i) => {
-      const overridden = mergeNoxConfigYAML(
-        baseNoxConfig,
-        computePeerConfig.nox ?? {},
-      );
+      const { rawConfig: computePeerRawNoxConfig, ...computePeerNoxConfig } =
+        computePeerConfig.nox ?? {};
+
+      let overridden = mergeNoxConfigYAML(baseNoxConfig, computePeerNoxConfig);
```
akim-bow, line 575:
```suggestion
      let overridenNoxConfig = mergeNoxConfigYAML(baseNoxConfig, computePeerNoxConfig);
```

src/lib/rust.ts
```diff
@@ -120,23 +120,44 @@ const ensureRust = async (): Promise<void> => {
...
+    return true;
+  }
 
-    await execPromise({
-      command: RUSTUP,
-      args: ["--version"],
-    });
+  if (process.env.PATH === undefined || process.env.HOME === undefined) {
+    return false;
+  }
```
akim-bow, line 130: probably you can try to extract home dir with `os.homedir` ?
Main advantage that it's simpler and works across platforms.

shamsartem: ok. used homedir, it's probably a bit simpler
but no real reason to care about cross platform cause I fail anyway if on windows and ask user to install rust by themselves

akim-bow: I see. But as far as i know we had some plans to support windows someday, right? If we use crossplatform API when possible it will make the window adoption a bit easier. Even if it is a single function right now.

akim-bow (approved):

PR #651, opened 2023-12-15
test: deal upd with new spell

vasd85:


src/lib/helpers/utils.ts
```diff
@@ -54,7 +54,7 @@ export function stringifyUnknown(unknown: unknown): string {
     }
 
     if (unknown instanceof Error) {
-      return jsonStringify(unknown, Object.getOwnPropertyNames(unknown));
+      return `${unknown.message}\n${unknown.stack ?? ""}`;
     }
```
akim-bow, line 58: What is the reason for this change?
For example Error object could have more properties than just `message` and `stacktrace`

test/const.ts
akim-bow, line 1: I would suggest renaming `const.ts` -> `constants.ts`.
Sounds more familiar and clearer.
Up to you.

src/lib/helpers/utils.ts
shamsartem: The reason is that when the error object is stringified like that it's rendered completely unreadable

I decided improve that according to your recommendation so other properties can be still visible, take a look

shamsartem (approved):

PR #652, opened 2023-12-15
fix: Support fire and forget (DXJ-562)

akim-bow:


src/commands/run.ts
```diff
@@ -489,11 +489,17 @@ const fluenceRun = async (args: RunArgs) => {
 
   await initFluenceClient(args, args.maybeFluenceConfig);
 
+  const def = functionCall.funcDef;
+
+  const returnTypeVoid =
+    def.arrow.codomain.tag === "nil" || def.arrow.codomain.items.length === 0;
```
shamsartem, line 495:
```suggestion
  const { codomain } = functionCall.funcDef.arrow;
  const returnTypeVoid = codomain.tag === "nil" || codomain.items.length === 0;
```

package.json
```diff
@@ -58,7 +58,7 @@
     "@fluencelabs/aqua-to-js": "0.3.0",
     "@fluencelabs/deal-aurora": "0.2.22",
     "@fluencelabs/fluence-network-environment": "1.1.2",
-    "@fluencelabs/js-client": "0.5.3",
+    "@fluencelabs/js-client": "0.5.4-DXJ-562-6db9e47-2164-1",
```
shamsartem, line 61: is js client with this not released yet?
please don't forget to fix this

shamsartem (approved):

PR #655, opened 2023-12-20
fix: service not added to default deal services

shamsartem:


src/lib/addService.ts
```diff
@@ -126,19 +126,14 @@ export const addService = async ({
...
   if (
-    isInteractive &&
-    fluenceConfig.deals !== undefined &&
-    DEFAULT_DEAL_NAME in fluenceConfig.deals &&
-    !(fluenceConfig.deals[DEFAULT_DEAL_NAME].services ?? []).includes(
-      serviceName,
-    ) &&
-    (interactive
-      ? await confirm({
+    isServiceMissingInDefaultDeal(fluenceConfig, serviceName) &&
```
folex, line 129: Could you pls add a comment on when this can be `true`? When a service can be in deal already on `addService`

shamsartem: after thinking about I confirmed that this should never happen because everything should be validated beforehand. But thanks to your concern I was able to improve ux a little bit in this regard

PR #661, opened 2023-12-21
test: dxj-573 deal upd with a new service

vasd85:


src/lib/configs/initConfig.ts
```diff
@@ -456,7 +456,7 @@ export function getReadonlyConfigInitFunction<
   };
 }
 
-const initializedConfigs = new Set<string>();
```
akim-bow, line 459: Why not just `configs` ?

test/sharedSteps.ts
```diff
@@ -187,10 +197,28 @@ export async function createSpellAndAddToDeal(
...
+      if (a.host_id < b.host_id) {
+        return -1;
+      }
+
+      if (a.host_id > b.host_id) {
+        return 1;
+      }
+
+      return 0;
+    })
```
akim-bow, line 212: I suggest to use lodash sort here.

akim-bow: `sortBy(result, 'host_id')`

shamsartem: The bad thing about lodash is that id doesn't care about type information so you might run into the problem because of that

src/lib/configs/initConfig.ts
shamsartem: it probably can be just `configs` but I suggest to leave it as it is, it's a map of the configs that were initialized previously, so here seems reasonable to mention this fact in the variable name

shamsartem (approved):

PR #665, opened 2023-12-26
test: DXJ-576 updating the deal by adding a module

vasd85:


test/schemasAndValidation.ts
akim-bow, line 1: I don't really like this filename.
I suggest to create a folder `validators` and add in there files according to each validator, e.g. worker validator, service validator.

test/sharedSteps.ts
```diff
@@ -61,6 +102,50 @@ export async function getFluenceConfig(cwd: string) {
...
+  const pathToServiceDir = getServiceDirPath(cwd, serviceName);
+
+  const serviceConfig = await initServiceConfig(
+    relative(cwd, pathToServiceDir),
+    cwd,
+  );
+
+  assert(
+    serviceConfig !== null,
+    `we create a service at ${pathToServiceDir} above - so the config is expected to exist`,
```
akim-bow, line 115:
```suggestion
    `the service, owning this config, must be created at the path ${pathToServiceDir}`,
```

test/sharedSteps.ts
```diff
@@ -255,65 +340,44 @@ export async function waitUntilShowSubnetReturnsExpected(
...
-        Object.prototype.hasOwnProperty.call(config.services, serviceName),
-      `${serviceName} is expected to be in services property of ${config.$getPath()} after ${serviceName} is added to it`,
-    );
+      const runDeployedServicesResult = JSON.parse(result);
 
-    return config;
-  };
+      if (!validateDeployedServicesAnswerSchema(runDeployedServicesResult)) {
+        throw new Error(
+          `result of running ${RUN_DEPLOYED_SERVICES_FUNCTION_CALL} aqua function is expected to be an array of Answer, but it is: ${result}`,
```
akim-bow, line 362:
```suggestion
          `result of running ${RUN_DEPLOYED_SERVICES_FUNCTION_CALL} aqua function is expected to be an array of Answer, but actual result is: ${result}`,
```

shamsartem (approved):
looks fine in general, I think

PR #671, opened 2023-12-27
feat(cli): Add header to default spell aqua

InversionSpaces:


src/lib/const.ts
```diff
@@ -579,7 +579,12 @@ func getInfos(peers: []PeerId) -> []Info:
 `;
 };
 
-export const SPELL_AQUA_FILE_CONTENT = `import Op, Debug from "${AQUA_LIB_NPM_DEPENDENCY}/builtin.aqua"
+export const SPELL_AQUA_FILE_CONTENT = `aqua DefaultSpell
```
folex, line 582: I wonder if it should take module name from the spell name

shamsartem (approved):

PR #688, opened 2024-01-03
feat: Intro CLI for Windows [fixes FLU-610]

akim-bow:
In this PR i'll try to put CLI in the working state in Windows system.
- Deal deploy should work
- Aqua compilation should work
- CLI should install on windows correctly

However some non critical features would be omitted
- Update won't work in windows yet (need to do some additional job, will do in separate PR)


src/lib/configs/project/service.ts
```diff
@@ -205,6 +206,10 @@ const getDefault: (
   relativePathToFacade: string,
   name: string,
 ): GetDefaultConfig => {
+  // All paths in config should be kept in platform independent way for consistency
+  // Also yaml parser would complain about backslashes, so it's better to avoid them instead of shielding
```
shamsartem, line 210: Good idea to keep these paths posix everywhere. Keep in mind that there are multiple places where the same problem happens (cli accepts paths in different places)
probably good idea to have a simple function that converts a path to posix and use it everywhere

README.md
shamsartem, line 1: I would suggest to not commit this file until we release on windows and everything works

install.ps1
shamsartem, line 1: I am not in favor of commiting stuff that is unused (if I am understanding correctly you don't plan to use this for now). If you still want to keep it (just in case) I would sugggest to create a branch in the cli repo with this file so we don't lose it

.github/workflows/release.yml
```diff
@@ -43,6 +43,8 @@ jobs:
           - linux-x64
           - darwin-x64
           - darwin-arm64
+          - win32-x64
+          - win32-x86
```
nahsi, line 47: We don't have windows runners :(

src/lib/helpers/package.ts
```diff
@@ -174,31 +175,31 @@ export const handleInstallation = async ({
...
-    } catch {
-      await installAndMoveToDependencyPath();
+      if (!isDependencyOverwrite) {
+        commandObj.error(
+          `Cannot finish installation. Reason: dependency ${name}@${version} already installed`,
+        );
+      }
     }
+
+    await rm(dependencyDirPath, { recursive: true, force: true });
```
shamsartem, line 198: don't really understand why you decided to make all these changes. Please explain. Why need to confirm stuff? Why you removed dependencyTmpDirPath (but for some reason not entirely as well, cause it still remains in type definition). I would not want to remove dependencyTmpDirPath cause I added that for the case when user starts installing something and doesn't wait till the end and excists the terminal. That happened several times and result is that some file is there but it didn't download completely so we it doesn't work (even though it can be accessed) So it's a bit more reliable to first download/install and then move file which is a very quick thing to do

src/lib/init.ts
```diff
@@ -280,11 +280,7 @@ const shouldInit = async (projectPath: string): Promise<boolean> => {
     message: `Directory ${color.yellow(projectPath)} is not empty. Proceed?`,
   });
 
-  if (hasUserConfirmedInitInNonEmptyDir) {
-    return true;
-  }
-
-  return false;
+  return hasUserConfirmedInitInNonEmptyDir;
```
shamsartem, line 283: nice catch

src/lib/rust.ts
```diff
@@ -188,9 +189,9 @@ const hasRequiredRustTarget = async (): Promise<boolean> => {
   return (
     await execPromise({
       command: RUSTUP,
-      args: ["target", "list"],
+      args: ["target", "list", "--installed"],
     })
-  ).includes(`${RUST_WASM32_WASI_TARGET} (installed)`);
+  ).includes(RUST_WASM32_WASI_TARGET);
```
shamsartem, line 194: this seems better in general as well. nice

src/lib/helpers/package.ts
akim-bow: Regularly moving, force-recursive deleting and creating directories isn't very quick thing to do, it takes dozens of ms to do that.

akim-bow: Nonetheless, i agree with your point. Will revert most of these changes.

src/beforeBuild.ts
```diff
@@ -144,26 +144,53 @@ const BIN_FILE_PATH = join(
...
+const WIN_BIN_FILE_PATH = join(
+  NODE_MODULES_DIR_NAME,
+  "oclif",
+  "lib",
+  "commands",
+  "pack",
+  "win.js",
+);
+
+async function insertContentNextLine(
```
shamsartem, line 156: I would suggest probably better name that explains what this function does
`patchOclif`

src/commands/service/repl.ts
```diff
@@ -147,13 +147,12 @@ Call ${facadeModuleConfig.name} service functions in repl like this:
...
 ${color.yellow(
   `call ${facadeModuleConfig.name} <function_name> [<arg1>, <arg2>]`,
-)}${SEPARATOR}}`);
+)}${SEPARATOR}`);
 
     await haltCountly();
 
     spawn(mreplPath, [fluenceTmpConfigTomlPath], {
       stdio: "inherit",
-      detached: true,
```
shamsartem, line 156: let's probably replace this with e.g. `!commandObj.config.windows`, cause on linux it works nice as it is

src/lib/rust.ts
```diff
@@ -215,8 +216,8 @@ type InstallCargoDependencyArg = {
   toolchain: string | undefined;
   name: string;
   version: string;
-  dependencyTmpDirPath: string;
   dependencyDirPath: string;
+  dependencyTmpDirPath: string;
```
shamsartem, line 220: please move it back, just to remove additional change that does nothing. Thanks!

src/commands/service/repl.ts
akim-bow: I also tested that on linux and haven't noticed any difference. So i'm curious why did you initially made that change?

shamsartem: I think the idea behind it was for the telemetry we send to Countly when command ends execution. but without detach command never ends execution. It's totally minor, but still

shamsartem: seems like I am mistaken cause in the code I can see
`await haltCountly();` before this runs
maybe it doesn't matter then
Probably fine to remove this detached if it doesn't really affect anything anyway

maybe the main idea was to just end nodejs process cause it's not needed anymore, it probably hangs in the background without detached

akim-bow: Reading the [doc ](https://nodejs.org/api/child_process.htm)right now and it says that:

- By default, the parent will wait for the detached child to exit. To prevent the parent from waiting for a given subprocess to exit, use the subprocess.unref() method.

- When using the detached option to start a long-running process, the process will not stay running in the background after the parent exits unless it is provided with a stdio configuration that is not connected to the parent. If the parent's stdio is inherited, the child will remain attached to the controlling terminal.

Seems like we need to do additional work/research to achieve desirable result. Mind if we do that in separate PR?

shamsartem: sure. nice. I really don't remember why I did it. There was some reason definetly, cause it's a weird thing to add

shamsartem (approved):

PR #700, opened 2024-01-09
feat: ensure memory limit of worker services is less then 2GB and pass it to installation spell [fixes DXJ-579 DXJ-545 DXJ-580 DXJ-546]

shamsartem:


src/lib/deployWorkers.ts
```diff
@@ -765,8 +773,12 @@ async function resolveWorker({
...
   );
 
-  const services: Upload_deployArgConfig["workers"][number]["config"]["services"] =
-    (workerConfig.services ?? []).map((serviceName) => {
+  const servicesWithUnresolvedMemoryLimit = (workerConfig.services ?? []).map(
+    (
+      serviceName,
+    ): Upload_deployArgConfig["workers"][number]["config"]["services"][number] & {
+      total_memory_limit: string | undefined;
+    } => {
```
akim-bow, line 781: Let's have a separate type for this. Probably even in this file above it there is no need to reuse it

src/lib/deployWorkers.ts
```diff
@@ -778,7 +790,8 @@ async function resolveWorker({
         )}`,
       );
 
-      const { overrideModules, serviceConfig } = maybeServiceConfig;
+      const { overrideModules, serviceConfig, totalMemoryLimit } =
+        maybeServiceConfig;
```
akim-bow, line 794:
```suggestion
        serviceConfig;
```
I still think that it's better to let ts config decide whether it's maybe config or no maybe. I don't feel like it's a good practise to put type related things in var name

src/lib/const.ts
```diff
@@ -55,6 +56,35 @@ export const defaultNumberProperties: Record<NumberProperty, number> = {
...
+  "giB",
+  "GiB",
+  "GIB",
+];
+
+export const BYTES_PATTERN = `^\\d+(\\.\\d+)?(\\s?)(${byteUnits.join("|")})$`;
+export const BYTES_FORMAT = `[number][whitespace?][B] where ? is an optional field and B is one of the following: ${byteUnits.join(
+  ", ",
+)}`;
+export const MAX_HEAP_SIZE_DESCRIPTION = `Max size of the heap that a module can allocate in format: ${BYTES_FORMAT}`;
```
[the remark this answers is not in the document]
shamsartem: It's very hard to say in js when something deserves a separate file. Probably it's good to have a lot of small files in a sense that it's possible to contain imports sometimes and not Node.js will not have to parse a bunch of stuff it will not use. But alternatively there can become too much of the files. and pretty messy dependency tree. So it's really hard to tell. For cli I have const file where all the consts live. Maybe it can be split somehow, and it will increase number of imports as well. So I am not really sure about this issue, more of a matter of taste

src/lib/deployWorkers.ts
shamsartem: I more or less try to remove these maybe things where they are not required, though there a benifits to have them sometimes. But right here not so straightforward, cause there is actually a serviceConfig inside maybeServiceConfig, where serviceConfig in the actual config from file and maybeServiceConfig in this case is just a thing that contains the actual config + overrides for modules + memory limit. Don't really kno,w how to rename that. We can leave it as it is or think of some why to rename stuff here, can't come up with a satisfying name

akim-bow: Maybe `combinedConfig` or `configWithOverrides` ?

akim-bow (approved):

PR #713, opened 2024-01-15
feat: improve docs and validation for memory limit [fixes DXJ-619 DXJ-546 DXJ-620]

shamsartem:


src/lib/configs/project/fluence.ts
```diff
@@ -268,6 +270,14 @@ const spellSchema: JSONSchemaType<FluenceConfigSpell> = {
...
   type: "object",
   properties: {
+    computeUnits: {
+      type: "number",
+      // These is the correct schema but Ajv thinks type here is incorrect
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+      const: 1 as unknown as null,
+      description: `Number of compute units you require. 1 compute unit = ${COMPUTE_UNIT_MEMORY_STR}. Currently the only allowed value is 1. This will change in the future. Default: 1`,
+      nullable: true,
+    },
```
akim-bow, line 280: Why not to use the number type here and add runtime validation? It would help with the `// eslint-disable...`. Also you would need to do less changes later, bcs schema type would be the same.

akim-bow (approved):

src/lib/configs/project/fluence.ts
shamsartem: Found an acceptable way to work around that for now

PR #719, opened 2024-01-16
fix(tests): Add `aqua` header to `smoke.aqua`

InversionSpaces:


shamsartem (approved):
Just a couple of them left :) Thanks!

PR #720, opened 2024-01-18
fix: override libp2p/utils version [fixes DXJ-629]

shamsartem:


shamsartem:
Fixed in js-client. No longer needed

PR #721, opened 2024-01-18
feat: simplify dependency related commands and rename dependency related properties in fluence.yaml [fixes DXJ-634 DXJ-598]

shamsartem:


src/commands/dep/install.ts
```diff
@@ -14,44 +14,59 @@
...
+      }
+
+      await ensureMarineAndMreplDependencies();
+      return;
     }
 
-    await installAllCargoDependencies({
-      maybeFluenceConfig,
-      force: flags.force,
-    });
```
akim-bow, line 53: What about installing cargo deps when `fluence dep install` is ran?

akim-bow: As far as i understand you have decided to drop support for cargo deps except marine and mrepl

shamsartem: it's done here https://github.com/fluencelabs/cli/pull/721/files#diff-404128a17fc921589b2109730c9a0997de53a1a2a1aa5ffc65969f8624077d90R57

also you don't really need to do that because they will be downloaded if they are missing anyway

shamsartem: other cargo dependencies were not really supported ever anyway. And I doubt we will have much more of them

src/lib/rust.ts
```diff
@@ -328,130 +317,178 @@ const tryDownloadingBinary = async ({
...
+
+export async function resolveMarineAndMreplDependencies() {
+  const fluenceConfig = await initFluenceConfig();
+
+  return [
+    [
+      "marine",
+      fluenceConfig?.marineVersion === undefined
+        ? versions.cargo.marine
+        : fluenceConfig.marineVersion,
```
akim-bow, line 420:
```suggestion
      fluenceConfig?.marineVersion ?? versions.cargo.marine,
```

akim-bow: The same thing below.
Not completely sure, but it looks like it's the usecase for null coalescing operator

shamsartem: yeah, just had something more complex there before, didn't realize thing are actually simpler

src/lib/npm.ts
```diff
@@ -130,90 +128,202 @@ export async function runNpm(args: Omit<ExecPromiseArg, "command">) {
...
+
+  stopSpinner();
+
+  if (aquaDependencies === null) {
+    throw new Error(
+      `Dependency was installed but for some reason wasn't able to get 'dependencies' property at ${color.yellow(
+        packageJSONPath,
+      )}. Please fix it manually. E.g. by removing ${color.yellow(
+        packageJSONPath,
+      )} and running install command again`,
```
akim-bow, line 185:
```suggestion
      `Dependency was installed but for some reason wasn't able to get 'dependencies' property at ${color.yellow(
        packageJSONPath,
      )}. Please fix it manually, e.g. by removing ${color.yellow(
        packageJSONPath,
      )} and running install command again`,
```

shamsartem: why? seems ok as it is

src/lib/npm.ts
```diff
@@ -130,90 +128,202 @@ export async function runNpm(args: Omit<ExecPromiseArg, "command">) {
...
 
-  fluenceConfig.dependencies.npm[packageName] = version;
-  await fluenceConfig.$commit();
+  if (
+    !dependenciesWithVersions.every((a): a is [string, string] => {
+      return typeof a[1] === "string";
+    })
+  ) {
+    return null;
+  }
```
akim-bow, line 266: You do a lot of manual validation here.
It's great that you validate it, but the code looks too verbose and not easy to read. I think it would be convenient and concise if we would have some kind of validation schema.
I suggest to use the schemas everywhere except very simple cases at it gives better readability and nearly instantly explains what's the code about.

src/lib/rust.ts
```diff
@@ -328,130 +317,178 @@ const tryDownloadingBinary = async ({
...
+  dependencyDirPath,
+  dependencyTmpDirPath,
+  name,
+  version,
+  toolchain,
+}: HandleInstallationArg): Promise<void> {
+  try {
+    // if dependency is already installed it will be there
+    // so there is no need to install
+    await access(dependencyDirPath);
```
akim-bow, line 463: Do you need add `return` statement here if dependency is installed and that's it ?

shamsartem: I think no real reason to add it cause this function doesn't contain anything else anyway and will exit on it's own
Only if some error happens - then the code in the catch will run

src/lib/npm.ts
```diff
@@ -130,90 +128,202 @@ export async function runNpm(args: Omit<ExecPromiseArg, "command">) {
...
+    return version;
+  }
+
+  let [pathA, pathB] = [projectRootDir, fluenceAquaDependenciesPath];
+
+  if (isBackToFluenceYAML) {
+    [pathA, pathB] = [pathB, pathA];
+  }
+
+  return `file:${relative(pathB, resolve(pathA, filePath))}`;
```
akim-bow, line 314: I'm not sure how this part does work. Can you put a couple of tests for that and/or the explanation? As far as i understand it, the code does the following:
- it gives the relative path from `.fluence/aqua-deps` to fluence project dir plus initial relative path.
.e.g. if i have following structure
`/fluence/src/aqua1/package.json`
`/fluence/src/aqua2/package.json`
and the first one has import `file:../aqua2`
The resulting path would be `file:../../../aqua2`
How will it help to find appropriate package?

shamsartem: You understand everything correctly. It's just transforming one relative path to another. I will add a good explanation for this cause it can seem confusing I agree

src/lib/rust.ts
akim-bow: Yes, you're right, sorry. Just misread the code.

akim-bow (approved):

PR #731, opened 2024-01-22
fix(tests): DXJ-617 fix environment problems

vasd85:
- environment issues are fixed
- some problems in tests are fixed
- made a refactoring


jest.config.js
```diff
@@ -14,8 +14,12 @@
  * limitations under the License.
  */
 
-/** @type {import('ts-jest').JestConfigWithTsJest} */
+import { RUN_TESTS_IN_PARALLEL } from "./dist/lib/setupEnvironment.js";
+
+/** @type {import("ts-jest").JestConfigWithTsJest} */
 export default {
+  maxConcurrency: process.env[RUN_TESTS_IN_PARALLEL] === "false" ? 1 : 1, // increasing number of concurrent tests leads to a problem with nonce
```
shamsartem, line 21: just use `1` here if that's what you wanna have

package.json
```diff
@@ -102,6 +102,7 @@
     "@types/lodash-es": "4.17.12",
     "@types/node": "20.11.0",
     "@types/platform": "1.3.6",
+    "@types/proper-lockfile": "^4.1.4",
```
shamsartem, line 105: please remove `^` we decided to use it like this cause it's a little bit more stable without it
```suggestion
    "@types/proper-lockfile": "4.1.4",
```

src/lib/helpers/utils.ts
```diff
@@ -53,6 +55,10 @@ export function stringifyUnknown(unknown: unknown): string {
       return String(unknown);
     }
 
+    if (unknown instanceof AssertionError) {
+      return String(unknown);
+    }
```
shamsartem, line 60: Replace thing above with
```suggestion
    if (unknown instanceof CLIError || unknown instanceof AssertionError) {
      return String(unknown);
    }
```

src/lib/setupEnvironment.ts
```diff
@@ -78,6 +78,6 @@ const isFluenceEnvWithoutCustom = getIsStringUnion(
 
 setEnvVariable(FLUENCE_ENV, isFluenceEnvWithoutCustom, "local");
 setEnvVariable(DEBUG_COUNTLY, isTrueOrFalseString, "false");
-setEnvVariable(RUN_TESTS_IN_PARALLEL, isTrueOrFalseString, "false");
+setEnvVariable(RUN_TESTS_IN_PARALLEL, isTrueOrFalseString, "true");
```
shamsartem, line 81: did you check both ways and setting this to true makes it run faster?
can you give any estimates for time it takes to run in both cases on your computer at least

shamsartem: or we can try it in CI whcih is more relevant

shamsartem: cause if difference is insignifact, as it was before - it would be nicer not to run in parallel cause at least logs would make sense in that case

test/helpers/localNetAccounts.ts
```diff
@@ -0,0 +1,192 @@
...
+      "0xbb3457514f768615c8bc4061c7e47f817c8a570c5c3537479639d4fad052a98a",
+  },
+  {
+    address: "0x6f10E8209296Ea9e556f80b0Ff545D8175F271d0",
+    privateKey:
+      "0xfbd9e512cc1b62db1ca689737c110afa9a3799e1bc04bf12c1c34ac39e0e2dd5",
+  },
+];
+
+async function processFile(
```
shamsartem, line 129: could you please explain the reasoning behind this?
would also be nice to have a doc comment for this function that roughly explains why it is needed, what kind of problem it solves

test/helpers/localNetAccounts.ts
```diff
@@ -0,0 +1,192 @@
...
+  }
+}
+
+function processData(data: string): number {
+  let index: number;
+
+  if (data === "") {
+    index = 0;
+  } else {
+    index = Number(data) + 1;
```
shamsartem, line 169: looks a bit weird. Why +1?

akim-bow: Probably you could use more meaningful function name bcs `processData` is too generic.

test/helpers/sharedSteps.ts
```diff
@@ -15,94 +15,100 @@
...
-  }
+  template: Template,
+): Promise<void> => {
+  const templatePath = getInitializedTemplatePath(template);
 
-  return join(getServiceDirPath(cwd, serviceName), "modules", moduleName);
-}
+  try {
+    await rm(cwd, { recursive: true });
+  } catch {}
```
akim-bow, line 108:
```suggestion
  await rm(cwd, { recursive: true, force: true });
```

Acording to Node.JS [docs](https://nodejs.org/api/fs.html), the force option here will prevent throwing the error.

test/helpers/testWrapper.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { TEST_DEFAULT_SENDER_ACCOUNT } from "./localNetAccounts.js";
+
+export const maybeConcurrentTest = (...args: Parameters<typeof test>): void => {
+  console.log(
+    `Running "${args[0]}" with sender account: ${TEST_DEFAULT_SENDER_ACCOUNT.address}`,
+  );
+
+  test.concurrent(...args);
+};
```
akim-bow, line 25: I'm standing against this wrapper. Adding just `console.log` above each test is kinda repetitive, but much simpler and cleaner. If you want to add some logs before each test in the suite, i suggest to use `beforeEach` hook for that specific suit.
Also, why would you need to do `test.concurrent` ? Isn't `it.concurrent` working?

vasd85: You are right, it should be deleted.
There was another logic in that function. I left it for debug purpose while setting up distribution of private keys in parallel test launch.

`test` and `it` are the same. So it doesn't matter what do you use.

test/helpers/localNetAccounts.ts
vasd85: I moved this method in the `test/helpers/utils.ts` because it may be used in other places

test/helpers/localNetAccounts.ts
```diff
@@ -126,64 +125,48 @@ export const LOCAL_NET_DEFAULT_ACCOUNTS: Account[] = [
...
+ */
+function increaseIndex(dataFromFile: string): string {
   let index: number;
 
-  if (data === "") {
+  if (dataFromFile === "") {
     index = 0;
   } else {
-    index = Number(data) + 1;
+    index = Number(dataFromFile) + 1;
```
akim-bow, line 141: I think we still need some explanation here regarding `+ 1`

test/tests/integration.test.ts
```diff
@@ -14,39 +14,35 @@
...
-
-const multiaddrs = await getMultiaddrs();
+  initializeTemplate,
+  multiaddrs,
+  runAquaFunction,
+} from "../helpers/sharedSteps.js";
 
 describe("integration tests", () => {
-  maybeConcurrentTest("should work with minimal template", async () => {
+  test.concurrent("should work with minimal template", async () => {
```
shamsartem, line 31: the benifit of using maybeConcurentTest here is that you can turn it off with an env variable if you want to run tests sequentially to see the logs clearly. But I understand that here it's impossible to turn off easilly now. I am I correct? If so, does `maybeConcurentTest` cause any problems or something? Why change that?

test/helpers/localNetAccounts.ts
akim-bow: I guess file stored actual index plus 1 ? Probably we can mention that somewhere

akim-bow: Or why not to store index directly as it is?

test/tests/integration.test.ts
akim-bow: @shamsartem i proposed to do so [here](https://github.com/fluencelabs/cli/pull/731#discussion_r1462162235).
Seems like the test execution is still configurable via env var. Please look at 2 previous Dmitriy's commits - he managed to configure test concurrency via env. @vasd85 can you confirm that?

shamsartem: hm so it's just controled with jest config `maxWorkers`. If this works - very nice

vasd85: Yes, the env var still works. I moved this "switcher" in the `jest.config.js` - `maxWorkers: process.env[RUN_TESTS_IN_PARALLEL] === "false" ? 1 : 5,`. If we run with 1 worker there is no way to parallelize tests.

test/helpers/localNetAccounts.ts
vasd85: According the current logic, the file stores index that have already used by other process and for the new process we need to get the next one. 

I can add an explanation here, or change the logic somehow.

akim-bow: small explanation is fine

vasd85: @akim-bow take a look please, seems now it's clear. I edited method docs and added a comment near the line.

akim-bow:
Nicely done, thanks

akim-bow (approved):

PR #734, opened 2024-01-25
fix: maybe fix seed bug [fixes DXJ-639]

shamsartem:


test/helpers/sharedSteps.ts
```diff
@@ -69,44 +66,58 @@ import {
   getServiceDirPath,
   getSpellAquaPath,
   getSpellDirPath,
-  pathToTheTemplateWhereLocalEnvironmentIsSpunUp,
 } from "./paths.js";
 
-export const multiaddrs = await getMultiaddrs();
+let localMultiaddrs: Node[] | undefined;
```
akim-bow, line 71: Can't find the code which effectively updates this variable.

test/helpers/sharedSteps.ts
```diff
@@ -69,44 +66,55 @@ import {
...
-      : [];
-
-  return {
-    kras: krasnodar,
-    stage: stage,
-    testnet: testNet,
-    local,
-  }[fluenceEnv];
+let localMultiaddrs: Node[] = [];
+
```
akim-bow, line 72: Still don't understand the benefit from having this variable compared to previous variable. It will be updated only once, bcs update happens when `localMultiaddrs.length === 0` is met. And after the first update, this condition will be always false.

shamsartem: this variable is just a cache that prevents `fluence default peers local` to be callied more then it needs to be called

previously getMultiaddrs was always called when this file is imported even if the code doesn't need this at all

akim-bow (approved):

PR #750, opened 2024-02-12
fix: adapt gateway template to vercel deployment

akim-bow:


src/lib/init.ts
```diff
@@ -376,14 +375,38 @@ async function initTSorJSGatewayProject({
...
 
+  await ensureDir(join(getGatewaySrcPath(), "app"));
+  await ensureDir(join(getGatewayPath(), "api"));
+  await ensureDir(join(getGatewayPath(), "public"));
+
   await Promise.all([
     fluenceConfig.$commit(),
-    writeFile(indexFilePath, getGatewayIndexJsContent(isJS), FS_OPTIONS),
+    writeFile(
+      join(getGatewayPath(), "src", "app", `index.${isJS ? "js" : "ts"}`),
```
shamsartem, line 385: a bit cleaner to get the estension once and then use it in all of the cases where it is needed

src/lib/init.ts
```diff
@@ -539,143 +562,142 @@ function stringifyError(e${isJS ? "" : ": unknown"}) {
...
 
-${
-  isJS
-    ? ""
-    : "type runDeployedServicesResponseType = Static<typeof runDeployedServicesResponse>;"
-}
+  ${
+    isJS
+      ? ""
+      : "type showSubnetResponseType = Static<typeof showSubnetResponse>;"
```
shamsartem, line 644: I think it's more or less accepted convention to name types starting with an uppercase letter
```suggestion
      : "type ShowSubnetResponseType = Static<typeof showSubnetResponse>;"
```

src/lib/init.ts
```diff
@@ -790,6 +813,77 @@ function getGatewayPackageJSON(isJS: boolean) {
...
 
+function getGatewayServerless(isJS: boolean) {
+  return `${
+    isJS
+      ? ""
+      : 'import type { TypeBoxTypeProvider } from "@fastify/type-provider-typebox";\n'
+  }import fastify from "fastify";
+import dotenv from "dotenv";
+
+// Runtime dependencies required for this function. Vercel does import only direct listed dependencies.
```
shamsartem, line 824:
```suggestion
// Runtime dependencies required for this function. Vercel does import only directly listed dependencies.
```
Also this seems super weird and I am really suprised they require users to do such things

src/lib/init.ts
```diff
@@ -790,6 +813,77 @@ function getGatewayPackageJSON(isJS: boolean) {
...
+(() => [
+  import("@fluencelabs/js-client"),
+  // This import will fail in runtime
+  () => import("@fluencelabs/marine-worker"),
+  import("@fluencelabs/marine-js"),
+  // This imports will impact loading speed
```
shamsartem, line 830:
```suggestion
  // These imports will impact loading speed
```
Also I don't know whcih ones? all of them listed after that?

src/lib/init.ts
```diff
@@ -790,6 +813,77 @@ function getGatewayPackageJSON(isJS: boolean) {
...
+    isJS
+      ? ""
+      : 'import type { TypeBoxTypeProvider } from "@fastify/type-provider-typebox";\n'
+  }import fastify from "fastify";
+import dotenv from "dotenv";
+
+// Runtime dependencies required for this function. Vercel does import only direct listed dependencies.
+(() => [
+  import("@fluencelabs/js-client"),
+  // This import will fail in runtime
```
shamsartem, line 827: I don't know what it means. You mean it will fail if you don't list it, omit it from here. Or will it fail anyway

src/lib/init.ts
```diff
@@ -376,14 +375,38 @@ async function initTSorJSGatewayProject({
...
+    writeFile(
+      join(getGatewayPath(), "api", `serverless.${isJS ? "js" : "ts"}`),
+      getGatewayServerless(isJS),
+      FS_OPTIONS,
+    ),
+    writeFile(
+      join(getGatewayPath(), "src", `dev.${isJS ? "js" : "ts"}`),
+      getGatewayDev(isJS),
+      FS_OPTIONS,
+    ),
```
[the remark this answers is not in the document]
akim-bow: Dev file is used for local startup. It won't work in serverless environment. It's used in `npm start` command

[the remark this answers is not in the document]
shamsartem: npm start that was generated for me looks like this

"start": "node --no-warnings src/server.js"

src/lib/init.ts
akim-bow: I restructured these imports

shamsartem (approved):

PR #768, opened 2024-02-20
fix: Add template structure to README [fixes DXJ-714]

akim-bow:
Added content to simplify navigation in starter templates


src/lib/const.ts
```diff
@@ -687,6 +689,15 @@ func spell():
 
 const QUICKSTART_README = `# Fluence Quickstart Template
 
+## Content
+
+- Default Marine service - \`${join(SRC_DIR_NAME, SERVICES_DIR_NAME)}\`.
+- Basic aqua functions - \`${join(SRC_DIR_NAME, AQUA_DIR_NAME)}\`.
+- Fluence HTTP Gateway for proxying Aqua execution - \`${join(
+  SRC_DIR_NAME,
+  GATEWAY_DIR_NAME,
```
shamsartem, line 698: Instead of joining paths yourself, the easier to maintain option would be to use getServicesDir function and others from paths.ts. Also if this part of readme is supposed to be completely the same as the one bellow I would just reuse it instead of duplicating it

src/lib/init.ts
```diff
@@ -561,15 +561,18 @@ function stringifyError(e${isJS ? "" : ": unknown"}) {
...
 ${isJS ? "" : 'import { type FastifyInstance } from "fastify";'}
 
 import { helloWorld, helloWorldRemote, showSubnet, runDeployedServices } from "../compiled-aqua/main.js";
 
 const DEFAULT_ACCESS_TOKEN = "abcdefhi";
-const DEFAULT_PEER_PRIVATE_KEY = new Array(32).fill("a").join("");
+
+const DEFAULT_PEER_PRIVATE_KEY = Buffer.from(
+  (await KeyPair.randomEd25519()).toEd25519PrivateKey(),
+).toString("base64");
```
shamsartem, line 575: Do you really need this? is this for an example or something. It doesn't look so good. Maybe js-client api needs to be a bit more user friendly in this regard. Anyway, since you are using a random key, probably you can remove all that, I was under impression js-client uses random key by default so this argument can be ommited

akim-bow: I do this to be more explicit, e.g., to show current private peer in server logs.

This snippet doesn't look good, yes. Because i need to convert private key accepted by JS client to CLI format and backwards. User generates private key via CLI and copies it to env variable. Probably from that perspective it's kinda ok.

src/lib/const.ts
akim-bow: I cannot use functions like `getServicesDir` because they reference `projectRootDir` and it cause runtime error.
Agree with point about duplication though

shamsartem: maybe something is wrong with the order of operations. projectRootDir is set when initing the project

shamsartem: You should be able to use these functions but make sure you use them after projectRootDir is set

akim-bow: How exactly can i make sure that?

shamsartem: you would have to look at the code and see the order of stuff happening

shamsartem: and in the template you would use `relative(pathToReadme, pathToTheActualThings)`

akim-bow: It works if i put `getServicesDir` inside function. But when i use it in constant in top-level scope, it fails. I don't want to convert constants to functions for the sake of that constant

akim-bow: Seems like circular dependency between files because some constants aren't initialized

PR #772, opened 2024-02-21
feat: up dev and provider ux

shamsartem:


src/commands/provider/commitment-info.ts
```diff
@@ -0,0 +1,99 @@
...
+    case 3n:
+      return "Inactive";
+    case 4n:
+      return "Failed";
+    case 5n:
+      return "Removed";
+    default:
+      return "Unknown";
+  }
+}
```
akim-bow, line 99: Minor suggestion:
Let's convert that `switch` statement to array and access specific status string by index in `status` param?

src/lib/deployWorkers.ts
```diff
@@ -544,9 +533,64 @@ const validateWasmExist = async (
...
+      if (choices.length === 0) {
+        return "Please select at least one deployment";
+      }
+
+      return true;
+    },
+    oneChoiceMessage(choice) {
+      return `One deployment found at ${fluenceConfig.$getPath()}: ${color.yellow(
+        choice,
+      )}. Do you want to select it`;
```
akim-bow, line 580:
```suggestion
      )}. Do you want to select it?`;
```

src/lib/helpers/utils.ts
```diff
@@ -200,6 +201,11 @@ export async function setTryTimeout<T, U>(
       isTrying = false;
       return res;
     } catch (e) {
+      if (failCondition !== undefined && failCondition(e)) {
```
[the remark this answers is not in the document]
shamsartem: linter will not like this cause we have strict boolean rule. Better to be explicit in this things to not accidentally shoot yourself into the foot

PR #789, opened 2024-02-22
fix: Add warning message when updating on Windows

akim-bow:


src/beforeBuild.ts
```diff
@@ -206,16 +199,10 @@ await patchOclif(
...
-);
+// Turn off update on windows
 
-// Don't build tarballs when packing for Windows
 await patchOclif(
-  PROMOTE_COMMAND_FILE_PATH,
-  "promoteGzTarballs(target)",
-  `...(target.platform === '${WINDOWS_PLATFORM}' ? [] : [promoteGzTarballs(target)])`,
+  UPDATE_COMMAND_FILE_PATH,
+  "const { args, flags } = await this.parse(UpdateCommand);",
```
shamsartem, line 206: move this to constant so it's easier to maintain

src/beforeBuild.ts
```diff
@@ -206,16 +199,10 @@ await patchOclif(
...
+  `if (this.config.platform === '${WINDOWS_PLATFORM}') this.error('You cannot use 'update' command on Windows. You can find newer releases here - https://github.com/fluencelabs/cli/releases. Remember to uninstall older CLI before updating, otherwise you could encounter hard to debug issues.'); const { args, flags } = await this.parse(UpdateCommand);`,
```
shamsartem, line 207: A bit simpler
```suggestion
  `if (this.config.windows) this.error('You cannot use 'update' command on Windows. You can find newer releases here - https://github.com/fluencelabs/cli/releases. Make sure to uninstall your current version of Fluence CLI before installing a new one, otherwise you could encounter hard to debug issues.'); const { args, flags } = await this.parse(UpdateCommand);`,
```

PR #790, opened 2024-02-22
feat: provider ux updates and dev ux fixes

shamsartem:


src/lib/chain/offer.ts
```diff
@@ -203,22 +218,27 @@ export function offersInfoToString({
...
-effectors: TODO
-`;
-            },
-          )
+          .map(({ offerId, offerName, offerInfo, offerExplorerInfo }) => {
+            return `Offer: ${color.yellow(offerName)}\n${yamlDiffPatch(
+              "",
+              {},
+              JSON.parse(
+                // this is needed to convert all BigInts to string
```
akim-bow, line 226: Seems a bit hacky. Why not convert them by hand?

src/lib/chain/offer.ts
```diff
@@ -257,68 +276,153 @@ export async function updateOffers(flags: OffersArgs) {
...
+    offerId,
+    effectors,
+    offerInfo: { offerInfo, offerExplorerInfo } = {
+      offerInfo: undefined,
+      offerExplorerInfo: undefined,
+    },
+  } of offersToUpdate) {
+    if (offerInfo === undefined || offerExplorerInfo === undefined) {
+      commandObj.warn(
+        `Can't find offer info for offer with id ${offerId} in chain. Please check if the offer exists in chain and try again. If the offer doesn't exist you can create it using '${CLI_NAME} provider offer-create' command`,
```
akim-bow, line 290:
```suggestion
        `Can't find offer info for offer with id ${offerId} in chain. Please check whether the offer exists in chain and try again. If the offer doesn't exist you can create it using '${CLI_NAME} provider offer-create' command`,
```

src/lib/chain/offer.ts
```diff
@@ -340,72 +444,176 @@ async function resolveOffersFromProviderConfig(flags: OffersArgs) {
...
+      return { result: offer };
     },
   );
 
   if (notFoundOffers.length > 0) {
     commandObj.error(
-      `Offers not found in ${providerArtifactsConfig.$getPath()} 'offer' property: ${notFoundOffers.join(
-        ", ",
-      )}`,
+      `Offer${notFoundOffers.length === 1 ? "" : "s"} ${color.yellow(
```
akim-bow, line 570: Let's have additional util function for that, e.g. `pluralize`?
It will take word and number. And depending on passed number, the function will return singular or plural form of the word. 
examples: `pluralize("Offer", 1)// Offer`, `pluralize("Cosmos", 2)// Cosmoses`

akim-bow (approved):

src/lib/chain/offer.ts
shamsartem: It's just a lot of values to unpack, it's a lot less code if it's done automatically for you, but I will probably have to do it because there is a price there and it needs to have decimals

src/lib/chain/offer.ts
shamsartem: I think both variants are acceptable. I prefer "if" cause it's shorter, less to read, less formal. I don't like to be unnecessarily formal in communications with developers

src/lib/chain/offer.ts
shamsartem: yeah, I thought about that, maybe something like that can be used, added. Mostly I try to avoid these situations, but here I couldn't come up with an alternative sentence that sounds good and doesn't have this issue

src/lib/chain/offer.ts
akim-bow: I think that they convey different meaning despite they could be used interchangeably sometimes.
`Whether` is used when there are 2 alternatives and you basically underlining that. 
In you case, the alternatives - offer does exist or doesn't.
Then you use `if` to formalize condition (if **A** happends, you can do **B**). 
I think that this distinction is kinda beneficial for developer, because more clearly passes the text meaning instead of putting `if` everywhere.

src/lib/chain/offer.ts
akim-bow: It's probably not bad, though kinda repetitive. You're clearly doing serialization in a typesafe way. Probably we can create util function for that, which recursively converts bigint to string. Anyway, i'm ok with any approaches keeping in mind our strict time limits :)

src/lib/chain/offer.ts
shamsartem: I really think this can be either way, but if you really like it better like that I don't mind

PR #817, opened 2024-02-27
feat: update deal-ts-clients

shamsartem:


src/commands/local/up.ts
```diff
@@ -65,6 +74,13 @@ export default class Up extends BaseCommand<typeof Up> {
       default: true,
     }),
     ...DOCKER_COMPOSE_FLAGS,
+    reset: Flags.boolean({
+      description:
+        "Reset docker-compose.yaml to default, remove volumes, remove previous local deployments",
```
akim-bow, line 79:
```suggestion
        "Resets docker-compose.yaml to default, removes volumes and previous local deployments",
```

src/commands/local/up.ts
```diff
@@ -74,8 +90,51 @@ export default class Up extends BaseCommand<typeof Up> {
...
+        await dockerCompose({
+          args: ["down"],
+          flags: {
+            v: true,
+          },
+          printOutput: true,
+          options: {
+            cwd: dirPath,
+          },
+        });
```
akim-bow, line 113: Seems like useless nesting. Probably we can improve that a bit with a helper function `noop`?
```suggestion
        await dockerCompose({
          args: ["down"],
          flags: {
            v: true,
          },
          printOutput: true,
          options: {
            cwd: dirPath,
          },
        }).catch(noop);
```

akim-bow (approved):

src/commands/local/up.ts
shamsartem: I think this one is more of a matter of taste. There are a number of places in CLI code where such pattern is used. It's easy to change later if we decide to

PR #839, opened 2024-03-08
fix: Update docs README.md to include correct command link

abdulmalik97:
1. command docs reference link fixed. 
2. Prioir took to an empty url


shamsartem:
I am fixing this as part of this PR. These links should not lead to main but to the correct tag instead https://github.com/fluencelabs/cli/pull/875

PR #840, opened 2024-03-09
feat: add cc-remove and other provider and developer improvements [fixes DXJ-752]

shamsartem:
- enabled metrics for noxes by default like this
```
tokio_metrics_enabled = true
metrics_enabled = true
metrics_timer_resolution = "1 minute"
```
- added `bootstrap_nodes` for non-local environments
- assert provider is registered after offer is selected so you see a prompt to connect your wallet only after that
- improved message for the request to connect your wallet
- allow using both IDs and names for all CC related commands. Getting cc ids by name from provider.yaml should hopefully work now on dar. Previously you had to use ids to add collateral
- added `cc-remove` command
- if deal status is Ended - ask user if he wants to replace previously deployed deal with a new one

### DIFF
--- a/src/lib/deploy.ts
+++ b/src/lib/deploy.ts
@@ -166,7 +166,9 @@ export async function deployImpl(this: Deploy, cl: typeof Deploy) {
         ? !(await confirm({
             message: `You previously deployed ${color.yellow(
               workerName,
-            )}, but this deal is already ended. Do you want to create a new deal and overwrite the old one? (at ${workersConfig.$getPath()})\nPlease keep this deal id ${previouslyDeployedDeal.dealIdOriginal} to withdraw the money from it after state overwrite`,
+            )}, but this deal is already ended. Do you want to create a new deal and overwrite the old one? (at ${workersConfig.$getPath()})\nPlease keep this deal id ${
+              previouslyDeployedDeal.dealIdOriginal
+            } to withdraw the money from it after state overwrite`,
             default: true,
           }))
         : true;
--- a/src/lib/deploy.ts
+++ b/src/lib/deploy.ts
@@ -166,7 +166,7 @@ export async function deployImpl(this: Deploy, cl: typeof Deploy) {
         ? !(await confirm({
             message: `You previously deployed ${color.yellow(
               workerName,
-            )}, but this deal is already ended. Do you want to create a new deal and overwrite the old one? (at ${workersConfig.$getPath()})\nPlease keep this deal id ${deal.id} to withdraw the money from it after state overwrite`,
+            )}, but this deal is already ended. Do you want to create a new deal and overwrite the old one? (at ${workersConfig.$getPath()})\nPlease keep this deal id ${previouslyDeployedDeal.dealIdOriginal} to withdraw the money from it after state overwrite`,
             default: true,
           }))
         : true;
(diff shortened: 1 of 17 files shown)


src/lib/deploy.ts
```diff
@@ -148,7 +152,32 @@ export async function deployImpl(this: Deploy, cl: typeof Deploy) {
...
 
+    let isDealUpdate = previouslyDeployedDeal !== undefined;
+
     if (previouslyDeployedDeal !== undefined) {
+      const deal = readonlyDealClient.getDeal(
+        previouslyDeployedDeal.dealIdOriginal,
+      );
+
+      const status = await deal.getStatus();
+      const isDealEnded = status === 2n;
```
folex, line 163: maybe convert it first and compare with a string? we sometimes change enum order, and it will break this code without anyone knowing...

src/lib/deploy.ts
```diff
@@ -148,7 +152,32 @@ export async function deployImpl(this: Deploy, cl: typeof Deploy) {
...
+
+      isDealUpdate = isDealEnded
+        ? !(await confirm({
+            message: `You previously deployed ${color.yellow(
+              workerName,
+            )}, but this deal is already ended. Do you want to create a new deal and overwrite an old one? (at ${workersConfig.$getPath()})\nYou will not be able to withdraw money from the old deal after that`,
```
folex, line 169:
```suggestion
            )}, but this deal is already ended. Do you want to create a new deal and overwrite the old one? (at ${workersConfig.$getPath()})\nPlease keep this dealId ${deal.id} to withdraw the money from it after state overwrite`,
```

folex (approved):

src/lib/deploy.ts
shamsartem: Yeah totally forgot to adjust that, ideally I would use enum exposed by deal-ts-client lib, but I think it's not exposed

src/lib/deploy.ts
```diff
@@ -148,7 +152,32 @@ export async function deployImpl(this: Deploy, cl: typeof Deploy) {
...
+      );
+
+      const status = await deal.getStatus();
+      const isDealEnded = status === 2n;
+
+      isDealUpdate = isDealEnded
+        ? !(await confirm({
+            message: `You previously deployed ${color.yellow(
+              workerName,
+            )}, but this deal is already ended. Do you want to create a new deal and overwrite the old one? (at ${workersConfig.$getPath()})\nPlease keep this deal id ${deal.id} to withdraw the money from it after state overwrite`,
```
shamsartem, line 169:
```suggestion
            )}, but this deal is already ended. Do you want to create a new deal and overwrite the old one? (at ${workersConfig.$getPath()})\nPlease keep this deal id ${previouslyDeployedDeal.dealIdOriginal} to withdraw the money from it after state overwrite`,
```

src/lib/deploy.ts
folex: it is exposed, I believe https://github.com/fluencelabs/deal/blob/117d49d9f9bb0dfb7e4b5efbff7fc03db1deeec5/ts-client/src/dealExplorerClient/types/schemes.ts#L282

shamsartem: The link leads to some type, but I need actual values, e.g. an enum

folex:
let's merge?

shamsartem:
> let's merge?

I am ok. I will improve enum stuff later
If you are gonna use it - I will have more time to fix stuff if something goes wrong if merge now

PR #849, opened 2024-03-11
feat: add ability for provider to update effectors and add module pack command

shamsartem:


src/lib/configs/project/module.ts
```diff
@@ -117,6 +119,11 @@ const overridableModulePropertiesV0 = {
     description:
       "A map of binary executable files that module is allowed to call. Example: curl: /usr/bin/curl",
   },
+  cid: {
+    description: "CID of the module when it was packed",
+    type: "string",
+    nullable: true,
+  },
```
kmd-fl, line 126: What about the import crate? Will the config be discarded as incorrect if I add it to the module.yaml?

shamsartem: I can add this optional property too. How do you suggest to call it?

kmd-fl: Dunno. Smth like "import-crate", I guess. I would rather not added it though, but whatever, we can delete it later, right? 
Also, what about adding CID to fluence.yaml (and provider.yaml for local deploy) on importing the module?

folex (approved):

PR #855, opened 2024-03-14
feat: new modules, gen Config.tomls for services, some refactoring, fix CID

shamsartem:


src/commands/module/build.ts
```diff
@@ -0,0 +1,80 @@
...
+  MARINE_BUILD_ARGS_FLAG_NAME,
+  MODULE_CONFIG_FULL_FILE_NAME,
+} from "../../lib/const.js";
+import { initCli } from "../../lib/lifeCycle.js";
+import { initMarineCli } from "../../lib/marineCli.js";
+import { input } from "../../lib/prompt.js";
+
+const PATH = "PATH";
+
+export default class Build extends BaseCommand<typeof Build> {
```
kmd-fl, line 37: It seems the `--no-input` flag suppresses output 
```
$ ~/Downloads/ll/fluence/bin/run.js module build --no-input effector
<no output>
$ ~/Downloads/ll/fluence/bin/run.js module build  effector
    Finished release [optimized] target(s) in 0.05s
```
fcli build: @fluencelabs/cli/0.15.25-new-modules-fd8363e-6427-1.0 from CI

shamsartem: yeah, seems this logic was there for a long time for all external binaries that CLI calls. Doesn't look very correct to me as well. Removed it

kmd-fl (approved):
Approve, but I checked locally only module build and module pack

PR #882, opened 2024-03-29
feat: set default price per epoch as 0.33 [fixes DXJ-757]

shamsartem:


src/lib/configs/project/fluence.ts
```diff
@@ -947,9 +945,9 @@ deployments:
   ${DEFAULT_DEPLOYMENT_NAME}:
     targetWorkers: ${TARGET_WORKERS_DEFAULT} # max amount of workers in the deal
     pricePerWorkerEpoch: "${DEFAULT_PRICE_PER_EPOCH_DEVELOPER}" # price per worker epoch in ${PT_SYMBOL}
-    initialBalance: "${DEFAULT_INITIAL_BALANCE}" # initial balance  after deploy in ${PT_SYMBOL}
     services: [] # list of service names to be deployed to this worker
     spells: [] # list of spell names to be deployed to this worker
+#   # initialBalance: "1.98"
```
justprosh, line 950: is it ok?

justprosh (approved):

src/lib/configs/project/fluence.ts
shamsartem: yeah it's just a comment that is used as an example. It will appear only for those who wanna see docs in the configs as comments (it's possible to turn this on in user's config.yaml by setting `docsInConfigs: true`)

justprosh: but why do you use # twice?

PR #885, opened 2024-04-02
feat: allow updating peers and CU in offer [DXJ-668]

shamsartem:
- offer update should work fully, so you can now also add and remove peers and compute units. Remember that offer update doesn't work after you create commitments
- filter out cc with status != WaitDelegation on cc-activate command. Fail if it happens that filtered out everything to avoid useless tx
- filter out cc with status != Inactive && status != Failed when withdrawing collateral
- local up will register all the offers in provider.yaml instead of only the first one and create commitments for all peers in all offers (this will help avoid the error when previously cc were created for all peers when it's possible for some of the peers to not belong to any offer)
- also you can now use cc-create with offer name to create cc for all peers in the offer. We probably could add this flag to other commands as I suggested previously, so you can operate on all nox-names belonging to one offer at the same time. 
- update deal-ts-clients
- print some info for all the users outside debug logs when retrying some operation so it's more clear that something is happening instead of the appearance of CLI hanging
- provider signing-wallet -> provider info that will also return peerids together with wallet address
add provider-secrets.yml to gitignore


justprosh (approved):

src/lib/chain/commitment.ts
```diff
@@ -111,7 +113,7 @@ export async function getComputePeersWithCC(
       c.commitmentCreatedEvent?.args.commitmentId ??
       c.marketGetComputePeerRes?.commitmentId;
 
-    if (commitmentId === undefined) {
+    if (commitmentId === undefined || commitmentId === ethers.ZeroHash) {
```
AlcibiadesCleinias, line 116: ZeroHash32

PR #888, opened 2024-04-04
feat: update eslint

shamsartem:


src/lib/configs/project/fluence.ts
```diff
@@ -984,7 +984,7 @@ ${yamlDiffPatch(
 # customFluenceEnv:
 #   contractsEnv: local
 #   relays:
-#     - "${kras[0]?.multiaddr}"
+#     - "/ip4/127.0.0.1/tcp/10010/ws/p2p/12D3KooWGpqaLpVUoLB5cJPLDS6CjBNZePBNTYPGpNPNZMDs6osY"
```
folex, line 987: why?

src/lib/helpers/utils.ts
```diff
@@ -66,7 +68,12 @@ export function jsonStringify(unknown: unknown): string {
 
 export function stringifyUnknown(unknown: unknown): string {
```
folex, line 69: Could you please add a comment on what's the use-case for this function and what cases this function covers. 

Seems like it's error oriented, maybe call it correspondingly?

folex (approved):
Awesome!

src/lib/configs/project/fluence.ts
shamsartem: this is just an example comment anyway that most people will not ever see cause we turned off comments in configs by default. It's just simpler to have an example that doesn't unnecesserally rely on anything concrete (such as here where kras[0] can be undefined according to typescript even though we know it never is)

src/lib/helpers/utils.ts
shamsartem: yeah I mostly use this with errors cause one can throw anything in js (not only errors) and it's good to be able to represent the error in a string form. but I also used it for e.g. debug logs or "unreachable error" messages where we don't neceserraly care much about the output as long as it's somewhat readable. Maybe would be good to have two functions for this each with a more clear purpose for existance and used accordingly. Added task to do that https://linear.app/fluence/issue/DXJ-763/refactor-stringifyunknown-function

PR #892, opened 2024-04-05
feat: decrease retry timeouts [fixes DXJ-764]

shamsartem:


src/lib/deal.ts
```diff
@@ -205,9 +205,11 @@ export async function match(dealAddress: string) {
       return dealMatcherClient.getMatchedOffersByDealId(dealAddress);
     },
     (err) => {
-      commandObj.error(stringifyUnknown(err));
+      commandObj.error(
+        `Wasn't able to find a match for the deal ${dealAddress}: ${stringifyUnknown(err)}`,
+      );
     },
-    1000 * 60 * 2, // 2 minutes
+    1000 * 5,
```
folex, line 212:
```suggestion
    1000 * 5 // 5 seconds,
```

src/lib/dealClient.ts
```diff
@@ -140,9 +140,11 @@ async function createDealClient(
       await core.minDealDepositedEpochs();
     },
     (err) => {
-      throw new Error(stringifyUnknown(err));
+      commandObj.error(
+        `Check if blockchain client is connected by running core.minDealDepositedEpochs() failed: ${stringifyUnknown(err)}`,
+      );
     },
-    1000 * 60 * 5,
+    1000 * 5,
```
folex, line 147:
```suggestion
    1000 * 5 // 5 seconds,
```

src/lib/dealClient.ts
```diff
@@ -255,11 +257,15 @@ export async function sign<T extends unknown[]>(
     (err) => {
       throw err;
     },
-    1000 * 60 * 3,
+    1000 * 5,
```
folex, line 260:
```suggestion
    1000 * 5, // 5 seconds
```

folex (approved):

src/lib/dealClient.ts
folex, line 1: What about number of retries while submitting a TX? How to control it?

shamsartem: retry happens every second so it most likely will be couple of them. If we are intersted specifically in trying fixed number of times - we should improve this setTryTimeout function

folex: Let's see how it goes. I just wasn't sure it applies only to `minDealDepositedEpochs` or to any transaction

PR #894, opened 2024-04-05
fix: use readonly chain client where currently possible [fixes DXJ-750]

shamsartem:


shamsartem:
> could you please list which commands now do not require MetaMask?
would take long time to do that and probably not super necesary to make a list
I would say that all the commands that actually require your secret key by definition will ask you to connect your wallet
The only exception is deal-list which just needs provider's public key and the most straightforward way to get the correct address is just to connect wallet. We can discuss the best way to improve this separately

vasd85:
The following commands work without signature:
```
provider cc-info
provider deal-rewards-info
provider info
provider offer-info
deal info
```

PR #897, opened 2024-04-05
fix: collateral withdraw by doing returnComputeUnitFromDeal and improve batch signing types [fixes DXJ-766]

shamsartem:


shamsartem:
fixed in https://github.com/fluencelabs/cli/pull/1025

PR #906, opened 2024-04-11
feat: offer-info updates [fixes DXJ-770]

shamsartem:


src/lib/configs/project/providerArtifacts.ts
```diff
@@ -151,18 +211,62 @@ const migrations: Migrations<Config> = [
...
+        throw new Error(
+          `Unreachable. Migration error. Unknown env ${env} in ${PROVIDER_ARTIFACTS_CONFIG_FULL_FILE_NAME}`,
+        );
+      }
+
+      const newConfigPerEnv: Record<string, OfferConfigV2> = {};
+
+      for (const [offerName, offer] of Object.entries(configPerEnv)) {
+        const providerAddress = await input({
+          message: `Enter provider address that was used when creating offer ${offerName} with offerId: ${offer.id} at ${configPath}`,
```
folex, line 246: Provider address is not stored along with the offers? Maybe it should be?

src/lib/chain/offer.ts
```diff
@@ -334,52 +327,51 @@ export async function updateOffers(flags: OffersArgs) {
     );
   }
 
+  const [notFoundOffersInfo, offersInfo] = await getOffersInfo(offersToUpdate);
+
+  if (notFoundOffersInfo.length > 0) {
+    commandObj.warn(
+      commandObj.warn(
+        `Can't find offer info using the indexer:\n${notFoundOffersInfo
```
folex, line 335: User won't know what is an "indexer".

```
Cannot find the following offers:
  - name: id
  - name: id
```

Maybe also print full indexer output in DEBUG in that case. But could be clumsy from the code organisation perspective.

folex (approved):

src/lib/configs/project/providerArtifacts.ts
shamsartem: this IS the idea in this PR to now store the provider address with the offer
here is the new config type https://github.com/fluencelabs/cli/pull/906/files/181239fe1dea4b3f430e2b79b630e7165016b77e#diff-75e8e297e5ce3647b9b327aa39919ad1bd3d7783560bfc361dba61658c04b15eR122-R125

PR #913, opened 2024-04-23
feat: make it possible to move noxes from one offer to another, improve tx batch types [fixes DXJ-759]

shamsartem:


src/commands/chain/proof.ts
```diff
@@ -20,7 +20,7 @@ import { BaseCommand, baseFlags } from "../../baseCommand.js";
 import { peerIdToUint8Array } from "../../lib/chain/conversions.js";
 import { setChainFlags, chainFlags } from "../../lib/chainFlags.js";
 import { CHAIN_FLAGS, PRIV_KEY_FLAG_NAME } from "../../lib/const.js";
-import { getDealClient, signBatch } from "../../lib/dealClient.js";
+import { getDealClient, signBatch, populate } from "../../lib/dealClient.js";
```
folex, line 23: `populate` sounds like a too generic of a name for a public API. No way to tell what it does by its name :(

shamsartem: maybe `populateTx`?

folex: I wonder where it comes from, cannot find it in the fluencelabs/deal 🤔 

Also, is it OK that you import `.js` and not typed `.ts`?

shamsartem: it comes from CLI. You can see it defined in this very PR.

using js extension is what I am doing for a long time. It's a eslint rule that enforced a while back cause there were some problems that I already do not remember that went away if you use it this way instead of using no-extension/ts-extsion

folex: Thank you for the explanation!

vasd85 (approved):

src/lib/chain/offer/updateOffers.ts
```diff
@@ -0,0 +1,456 @@
...
+  printOffersToUpdateInfo(populatedTxs);
+
+  if (
+    !(await confirm({
+      message: "Would you like to continue",
+      default: true,
+    }))
+  ) {
+    commandObj.logToStderr("Offers update canceled");
+    return;
```
folex, line 72: Will it return exit 1? I think it should 🤔

src/lib/chain/offer/updateOffers.ts
```diff
@@ -0,0 +1,456 @@
...
+  resolveOffersFromProviderConfig,
+  type EnsureOfferConfig,
+} from "./offer.js";
+
+export async function updateOffers(flags: OffersArgs) {
+  const offers = await resolveOffersFromProviderConfig(flags);
+  const offersFoundOnChain = filterOffersFoundOnChain(offers);
+  const populatedTxs = await populateUpdateOffersTxs(offersFoundOnChain);
+
+  const txsToSign = [
```
folex, line 45:
```suggestion
  const updateOffersTxs = [
```

src/lib/chain/offer/updateOffers.ts
```diff
@@ -0,0 +1,456 @@
...
+      message: "Would you like to continue",
+      default: true,
+    }))
+  ) {
+    commandObj.logToStderr("Offers update canceled");
+    return;
+  }
+
+  await assertProviderIsRegistered();
+  await signBatch(txsToSign);
```
folex, line 76: Shouldn't there be some error handling and offer-update specific reports?

src/lib/chain/offer/updateOffers.ts
```diff
@@ -0,0 +1,456 @@
...
+        .join(
+          "\n",
+        )}\n\nPlease make sure the offers exist. If the offers don't exist you can create them using '${CLI_NAME} provider offer-create' command`,
+    );
+  }
+
+  return offersToUpdateFoundOnChain;
+}
+
+type PeersOnChain = {
```
folex, line 118: I'd move all types to the top of the file, so it's easy to see the whole nomenclature / domain before reading the operations applied.

src/lib/chain/offer/updateOffers.ts
```diff
@@ -0,0 +1,456 @@
...
+        }),
+      )) satisfies PeersOnChain;
+
+      const effectorsOnChain = await Promise.all(
+        offerIndexerInfo.effectors.map(({ cid }) => {
+          return cidHexStringToBase32(cid);
+        }),
+      );
+
+      const txsToRunBeforeOthers = (await populatePeersToRemoveTxs(
```
folex, line 147: Why are these to run before others? What are "others"?

src/lib/chain/offer/updateOffers.ts
shamsartem: it will not. I don't think it should. It's normal operation. You wanted to update. You inspected the diff. You decided not to proceed yourself

src/lib/chain/offer/updateOffers.ts
folex, line 147: maybe?
```suggestion
      const removePeersFromOffersTxs = (await populatePeersToRemoveTxs(
```

src/lib/chain/offer/updateOffers.ts
```diff
@@ -0,0 +1,456 @@
...
+          populateEffectorsRemoveTx(offer, effectorsOnChain),
+          populateEffectorsAddTx(offer, effectorsOnChain),
+
+          populateCUToRemoveTxs(offer, peersOnChain),
+          populateCUToAddTxs(offer, peersOnChain),
+
+          populatePaymentTokenTx(offer),
+          populateMinPricePerWorkerEpochTx(offer),
+        ])
+      ).flat() satisfies Txs;
```
folex, line 165: awesome!

src/lib/chain/offer/updateOffers.ts
shamsartem: yeah because you have to first remove before adding. I will rename so it's clearer

src/lib/chain/offer/updateOffers.ts
shamsartem: can do

folex (approved):

PR #929, opened 2024-05-13
feat: chunk CUs when withdrawing collateral, add retry for Tendermint RPC error

shamsartem:


src/lib/chain/commitment.ts
```diff
@@ -457,10 +462,11 @@ export async function collateralWithdraw(flags: CCFlags) {
...
 
-    await signBatch([
-      populateTx(capacity.removeCUFromCC, commitmentId, [...unitIds]),
-      populateTx(capacity.finishCommitment, commitmentId),
-    ]);
+    for (const unitIdsBatch of chunk([...unitIds], flags[MAX_CUS_FLAG_NAME])) {
+      await sign(capacity.removeCUFromCC, commitmentId, [...unitIdsBatch]);
+    }
+
+    await sign(capacity.finishCommitment, commitmentId);
```
folex, line 469: Maybe if `unitIds.length < MAX_CUS`, then use `signBatch`? So that it's a single tx for the low amount of CUs.

And then catch the error, and print something like `"If you observe a problem with gas usage, try to pass a lower number to --max-cus"`

folex (approved):

folex:
Can you write a test on it btw?

shamsartem:
> Can you write a test on it btw?

https://linear.app/fluence/issue/DXJ-777/set-foundation-for-provider-tests

PR #977, opened 2024-07-09
fix: documentation links in configs

shamsartem:


folex (approved):
I wonder if we fix previous versions somehow... Maybe copy docs to both places? to `cli/docs/configs` and to `docs/configs`? Just for a few months.

PR #980, opened 2024-07-11
feat: add provider test

shamsartem:


cli/docs/commands/README.md
```diff
@@ -643,13 +643,14 @@ Withdraw FLT collateral from capacity commitment
...
   $ fluence delegator collateral-withdraw [IDS] [--no-input] [--env <dar | kras | stage | local | custom>] [--priv-key
-    <private-key>] [--max-cus <value>]
+    <private-key>] [--max-cus <value>] [--finish]
 
 ARGUMENTS
   IDS  Comma separated capacity commitment IDs
 
 FLAGS
   --env=<dar | kras | stage | local | custom>  Fluence Environment to use when running the command
+  --finish                                     Finish capacity commitment after collateral withdrawal
```
justprosh, line 653: What is the meaning of it? Why we need to withdraw collateral and not finish commitment?

cli/docs/commands/README.md
```diff
@@ -1081,8 +1083,11 @@ FLAGS
                                 local env 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 is used by
                                 default when CLI is used in non-interactive mode
       --quiet-pull              Pull without printing progress information
+      --[no-]set-up             Set up provider, offer, commitments and deposit collateral, so there is an active offer
```
justprosh, line 1086: Can you add comment about default behavior w/o this flag?

cli/docs/commands/README.md
shamsartem: it's for delegators (stackers) so they are not required to do that

justprosh: But stakers most likely will do that via UI and fcli will be used by providers that by default want to finish CC

cli/test/tests/provider.test.ts
```diff
@@ -0,0 +1,191 @@
...
+      await fluence({
+        args: ["provider", "cc-info"],
+        flags: {
+          ...PRIV_KEY_1,
+          [OFFER_FLAG_NAME]: NEW_OFFER_NAME,
+        },
+        cwd,
+      });
+
+      await sleepSeconds(CC_DURATION_MINUTES * 60);
```
justprosh, line 147: Wow, what do these sleeps mean? You actually wait so long in tests?

cli/docs/commands/README.md
shamsartem: yeah I see, can be a bit confusing. by default --set-up is set to true which not only runs docker container but also sets up provider and a working offer on the network. if you don't need it (e.g. you want to set up provider yourself using cli commands) you can pass a --no-set-up flag. Maybe it's better to just leave only a --no-set-up version cause --set-up doesn't make sence since it's already a default

cli/docs/commands/README.md
shamsartem: for providers CC is finished by default so it's fine

justprosh: Can you add it in description?

cli/docs/commands/README.md
justprosh: yeah, probably you need only --no-set-up

cli/test/tests/provider.test.ts
justprosh: It's better to mine blocks in anvil, but I think it requires more work right now

cli/docs/commands/README.md
shamsartem: where? this command is specifically for delegators so only they will see this flag. Providers don't have this flag and it just does the right thing for them. Delegators have a completely separate set of commands specifically for them (even though most probably they will not use CLI after all)

justprosh: Oh, sorry, I messed up, you're right

justprosh (approved):

PR #1019, opened 2024-09-04
feat: don't require connecting wallet on provider info if provider address is in provider artifacts for current env

shamsartem:


cli/src/lib/const.ts
```diff
@@ -425,6 +425,15 @@ export const PRIV_KEY_FLAG = {
   }),
 };
 
+export const ADDRESS_FLAG_NAME = "address";
+
+export const ADDRESS_FLAG = {
+  [ADDRESS_FLAG_NAME]: Flags.string({
+    description: "Account address",
```
folex, line 432:
```suggestion
    description: "Provider address",
```

folex (approved):

PR #1024, opened 2024-09-09
feat: match when not enough workers and running deal deploy, print success message in case of retry

shamsartem:


cli/src/lib/deploy.ts
```diff
@@ -338,6 +345,20 @@ export async function deployImpl(this: Deploy, cl: typeof Deploy) {
   await disconnectFluenceClient();
 }
 
+async function getDealStatus(dealId: string) {
+  const { DealStatus } = await import("@fluencelabs/deal-ts-clients");
+  const { readonlyDealClient } = await getReadonlyDealClient();
+  const deal = readonlyDealClient.getDeal(dealId);
+  const status = Number(await deal.getStatus());
+
+  function isDealStatus(status: number): status is DealStatus {
```
MixailE, line 354: Maybe we can move that function outside `getDealStatus`. What do you think about it?

cli/src/lib/deploy.ts
```diff
@@ -76,7 +78,12 @@ export const DEPLOY_FLAGS = {
   }),
 };
 
-type DealState = "notCreated" | "notMatched" | "matched" | "ended";
+type DealState =
```
MixailE, line 81: Maybe `enum` would be more preferable here?

cli/src/lib/deploy.ts
shamsartem: I put it inside cause this function is typescript-pleaser utility one-liner. No real need to use it or have it globally in the module. This should always return true except in the case if deal-ts-clients is not updated properly. If you don't mind I will leave it as it is cause it works totally fine either way in my opinion

cli/src/lib/deploy.ts
shamsartem: enum and union in this situation function exactly the same way from type-safety stand point except with enum you will have to add `DealState` prefix and import each time you want to check this state. I mostly prefer unions for this reason. Enums in my opinion are only appropriate if you have to put some weird constant behind the name (e.g. like we have with `DealStatus`)

MixailE (approved):
LGTM

PR #1031, opened 2024-09-12
feat: create one deal per peer with cuCountPerWorker number of compute units

shamsartem:


cli/src/lib/deal.ts
```diff
@@ -205,37 +205,29 @@ export async function createAndMatchDealsWithAllCUsOfPeerIds({
...
-  for (const { computeUnits, offerId, peerId } of offersWithCUs) {
-    for (const unitId of computeUnits) {
-      const dealAddress = await dealCreate({
-        ...dealCreateArgs,
-        maxWorkersPerProvider: 1,
-        targetWorkers: 1,
-        minWorkers: 1,
-      });
+  for (const { offerId, computeUnits, peerId } of offersWithCUs) {
+    const CUs = computeUnits.slice(0, dealCreateArgs.cuCountPerWorker);
```
justprosh, line 209: Are these CUs all available to deploy? So here is no CUs from this peer that participate it other deals?

shamsartem: Yeah, there is a filter statement above that I wrote before that filters out only CUs without deals

justprosh (approved):

Fedor4096 (approved):

PR #1033, opened 2024-09-13
feat: estimate gas before sending tx

shamsartem:


cli/src/lib/dealClient.ts
```diff
@@ -685,3 +685,64 @@ export async function batchRead<T>(rpcReadCalls: Array<() => Promise<T>>) {
...
+        signArgs.method,
+        ...getArgs(valuesToRegister),
+      ).populate();
+
+      populatedTx.from = address;
+      await providerOrWallet.estimateGas(populatedTx);
+
+      isValidTx = true;
+    } catch (e) {
+      sliceIndex = Math.floor(sliceIndex / 2);
```
crystalbit, line 729: maybe better `Math.ceil`? if we check bigger half, smaller half more probably will be ok

shamsartem: if I am not mistaken the number of transactions will be the same in either case

crystalbit: but if we floored passes, second part may fail if it has more count (+1). In case of ceil, second part may have same or -1 count

shamsartem: It will not fail. Rest of the parts are checked exactly the same way (there can be more then 2 if a lot of CUs it can be a lot of txs)

crystalbit: ok

crystalbit (approved):

PR #1053, opened 2024-09-19
feat!: use guessTxSizeAndSign in multicall

shamsartem:


cli/src/lib/chain/offer/offer.ts
```diff
@@ -315,68 +291,76 @@ ${await offersInfoToString([offerInfoErrors, offersInfo])}
...
-    offerName,
-  });
+  while (totalAddedCPs < allCPs.length) {
+    const remainingCPs = allCPs.slice(totalAddedCPs);
+
+    const CUsToRegisterCount = remainingCPs.flatMap(({ unitIds }) => {
+      return unitIds;
+    }).length;
+
+    const { registeredValues: addedCPs } = await guessTxSizeAndSign({
```
folex, line 320: can `addedCPs` ever be empty due to some error handling inside `guessTxSizeAndSign`?

I'm vary of this `while` becoming an infinite loop. Maybe it makes sense to exit with error if `addedCPs` is empty?

crystalbit (approved):

PR #1057, opened 2024-09-24
feat: add --reset-nox-secrets flag, improve config validation error messages [JS-812, JS-813]

shamsartem:


cli/src/commands/provider/gen.ts
```diff
@@ -15,29 +15,56 @@
...
 } from "../../lib/paths.js";
+import { confirm } from "../../lib/prompt.js";
 
 export default class Gen extends BaseCommand<typeof Gen> {
   static override description = `Generate Config.toml files according to ${PROVIDER_CONFIG_FULL_FILE_NAME} and secrets according to ${PROVIDER_SECRETS_CONFIG_FULL_FILE_NAME}`;
   static override examples = ["<%= config.bin %> <%= command.id %>"];
   static override flags = {
     ...baseFlags,
     ...CHAIN_FLAGS,
+    "reset-secrets": Flags.boolean({
```
folex, line 47: maybe `--reset-nox-secrets`? Otherwise sounds SCARY

cli/src/commands/provider/gen.ts
```diff
@@ -15,29 +15,56 @@
...
-    await initCli(this, await this.parse(Gen));
+    const { flags } = await initCli(this, await this.parse(Gen));
+    const providerSecretsConfigPath = await ensureProviderSecretsConfigPath();
+    const fluenceSecretsDir = getFluenceSecretsDir();
+
+    if (
+      flags["reset-secrets"] &&
+      (await confirm({
+        message: `Are you sure you want to remove ${color.yellow(providerSecretsConfigPath)} and ${color.yellow(fluenceSecretsDir)} to generate new ones`,
+        default: true,
```
MixailE, line 61: Maybe it should be `false`?

shamsartem: if default is false then there will be no way to run this non-interactively at all if it's really needed for some reason (e.g. in tests)

MixailE: Can we use env here?

shamsartem: how?

shamsartem: ah, also that "default" is only in regards of what happens in non-interactive mode by default. The actual default value that you can just press enter is still YES

cli/src/commands/provider/gen.ts
```diff
@@ -15,33 +15,128 @@
...
+      (await confirm({
+        message: `Are you sure you want to backup nox secrets ${color.yellow(providerSecretsConfigPath)} and ${color.yellow(fluenceSecretsDir)} (if they exist) to ${color.yellow(backupDirPath)} and generate new ones`,
+        default: flags[RESET_NOX_SECRETS_FLAG_NAME],
+      }))
+    ) {
+      if (
+        !flags[NO_WITHDRAW_FLAG_NAME] &&
+        (await confirm({
+          message:
+            "Do you want to withdraw remaining tokens from your noxes before continuing",
```
crystalbit, line 89: shall there be "?" in the end?

cli/src/lib/configs/initConfig.ts
```diff
@@ -415,38 +421,96 @@ export function getReadonlyConfigInitFunction<
...
+              ? ""
+              : `\n\nConfig version is ${color.yellow(currentConfigVersion)}, while the latest config version (that ${CLI_NAME_FULL} of version ${commandObj.config.version} is aware of) is ${color.yellow(
+                  latestConfigVersion,
+                )}. ${CLI_NAME_FULL} will automatically migrate your config to the latest version when you fix validation errors (that are listed above) and run any command that uses this config`
+          }`,
+        );
+      }
+
+      if (Number(config.version) < migrations.length) {
+        ({ latestConfig, configString } = await migrateConfig({
```
crystalbit, line 496: why in braces?

shamsartem: this one is the only way to destructure-reassign in js https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

cli/src/commands/provider/gen.ts
shamsartem: ? is added automatically in the beggining of every prompt by the lib that is used in CLI so in CLI code ? is not needed to be written explicitly in any of the prompts

crystalbit (approved):

PR #1064, opened 2024-10-30
feat: refactor configs [JS-831]

shamsartem:


packages/cli/package/src/lib/lifeCycle.ts
```diff
@@ -115,22 +87,16 @@ export async function initCli<
...
   if (requiresFluenceProject) {
-    setEnvConfig(await initNewEnvConfig());
+    await initNewEnvConfig();
   } else {
     try {
       // ensure env config also when provider.yaml exists
       await access(getProviderConfigPath());
-      setEnvConfig(await initNewEnvConfig());
-    } catch {
-      const envConfig = await initEnvConfig();
```
MixailE, line 128: We don't need that anymore?

packages/cli/package/src/commands/local/up.ts
```diff
@@ -97,48 +95,35 @@ export default class Up extends BaseCommand<typeof Up> {
...
   async run(): Promise<void> {
     const env: FluenceEnv = "local";
-
-    try {
-      const envConfig = await initNewEnvConfig(env);
-      envConfig.fluenceEnv = env;
-      await envConfig.$commit();
-      setEnvConfig(envConfig);
-    } catch (e) {
-      this.error(
```
MixailE, line 107: We don't need catch errors here anymore?

packages/cli/package/src/commands/provider/cc-collateral-withdraw.ts
```diff
@@ -1,43 +0,0 @@
...
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
- */
-
-import { BaseCommand } from "../../baseCommand.js";
-import { collateralWithdraw } from "../../lib/chain/commitment.js";
-import { CHAIN_FLAGS, FLT_SYMBOL, CC_FLAGS } from "../../lib/const.js";
-import { initCli } from "../../lib/lifeCycle.js";
-
-// Deprecated alias for provider cc-finish
-export default class CCCollateralWithdraw extends BaseCommand<
```
MixailE, line 24: We don't need `cc-collateral-withdraw`?

packages/cli/package/src/lib/lifeCycle.ts
shamsartem: I added a global "cache" for all configs long ago, cause essentially when you init a config object you can reuse it for the duration of CLI running cause it is expected to move anywhere, so there is no need anymore to store it globally separately. Finally removed that

packages/cli/package/src/commands/local/up.ts
shamsartem: there previously was a check for env config that fluence.yaml exists. Now there is no such check cause e.g. providers also need env config, so there are no errors thrown there anymore, nothing to catch

packages/cli/package/src/commands/provider/cc-collateral-withdraw.ts
shamsartem: it still exists as an alias, it's an old naming, this file can be removed now after I updated oclif

MixailE (approved):
LGTM

PR #1073, opened 2024-12-07
Improve uint8ArrayToHex

lucia-w:
- Use a string enum to centralize CLI tags
- convert Uint8Array in uint8ArrayToHex

### DIFF
--- a/packages/common/src/index.ts
+++ b/packages/common/src/index.ts
@@ -57,26 +57,20 @@ export type TransactionPayload = {
   transactionData: TransactionRequest;
 };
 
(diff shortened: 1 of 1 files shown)


shamsartem:
Thanks for the suggestion, but in our codebase we favor unions instead of enums cause they are less verbose while at least in this case accomplishing exactly the same goals

Also toString is not allowed in this codebase cause it's not type-safe. There are type-safe alternatives for converting to strings for everything. This way it's not possible to accidentally convert something to string in the wrong way if types change

PR #1074, opened 2024-12-10
fix: replace wasm32-wasi target with wasm32-wasip1

zenixls2:
dependency: https://github.com/fluencelabs/marine/pull/439
needs to wait until the above pr been released to set correct version in the dependencies of `/packages/cli/package/package.json`

refer to the renaming:
https://doc.rust-lang.org/nightly/rustc/platform-support/wasm32-wasip1.html
currently in beta and nightly, the wasm32-wasi target has been removed.

Original issue:
```bash
fluence init hello-fluence
```
Just doesn't compile. The rust version in cli is nightly 2024.6.10, which already removed `wasm32-wasi` from the target.
Also notice that the rust version in `marine` is different from `cli`

### DIFF
--- a/packages/cli/package/src/lib/const.ts
+++ b/packages/cli/package/src/lib/const.ts
@@ -48,7 +48,7 @@ export const PT_SYMBOL = "USDC";
 
 export const MAX_TOKEN_AMOUNT_KEYWORD = "max";
 
-export const RUST_WASM32_WASI_TARGET = "wasm32-wasi";
+export const RUST_WASM32_WASI_TARGET = "wasm32-wasip1";
 
 export const DEFAULT_MARINE_BUILD_ARGS = `--release`;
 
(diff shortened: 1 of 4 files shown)


zenixls2:
Just want to mentioned you that since January the old wasm target will also be removed from stable rust toolchain. You'll eventually need to apply similar changes within next 2 months. The steps written in getting started page only applies to the developers who don't have rust installed before.

For developers who already have rust installed in their environment, they are forced to downgrade the rust version without useful error messages from the fluence-cli when they try the getting started steps.

@shamsartem

shamsartem:
> tioned you that since January the old wasm target will also be removed from stable rust toolchain. You'll eventually need to apply similar changes within next 2 months. The steps written in getting started page only applies to the developers who don't have rust installed before.

thanks. As I said all of this will be removed anyway so it doesn't matter at this moment

PR #1076, opened 2024-12-17
feat!: remove all developer functionality, add kubernetes manifest generation and deployment

shamsartem:


shamsartem:
Discussed with @folex to review this after the fact

PR #1083, opened 2024-12-25
feat: add balance keeper

shamsartem:


shamsartem:
This is already done in another commit

PR #1092, opened 2025-01-13
feat: suggest to deploy manifests when ip changed

shamsartem:


shamsartem:
Merging without review to enable testing

PR #1096, opened 2025-01-17
feat!: add datacenter and resource tests, print offer-info to stdout

shamsartem:


packages/cli/package/src/lib/dealClient.ts
```diff
@@ -122,6 +131,11 @@ export async function resolveDeployment() {
       return {
         ...DEPLOYMENTS[await ensureChainEnv()],
         ...envConfig?.deployment,
+        // TODO: REMOVE THIS!
```
MixailE, line 134: Do we need that?

shamsartem: yes for now, soon Valery will fix and I will remove it

packages/cli/package/test/tests/provider.test.ts
```diff
@@ -165,6 +207,452 @@ describe("provider tests", () => {
...
+      const offerInfoWithUpdatedResources = await fluence({
+        args: ["provider", "offer-info"],
+        flags: {
+          ...PRIV_KEY_1,
+          [OFFER_FLAG_NAME]: defaultOfferName,
+        },
+        cwd,
+      });
+
+      expect(offerInfoWithUpdatedResources).toEqual(
```
MixailE, line 651: Maybe we need to check rest of updated values?

```
const updatedResourceTypes = [
    BANDWIDTH_RESOURCE_METADATA_UPDATED,
    CPU_RESOURCE_NAME_UPDATED,
    RAM_RESOURCE_NAME_UPDATED,
    STORAGE_RESOURCE_NAME_UPDATED,
    IP_RESOURCE_NAME_UPDATED,
  ].forEach(resource => {
    expect(offerInfoWithUpdatedResources).toEqual(
      expect.stringContaining(resource.type),
    );
  });
```

MixailE (approved):
LGTM

packages/cli/package/test/tests/provider.test.ts
shamsartem: yeah, will add something like this. Thanks

PR #1100, opened 2025-01-29
fix: yaml-diff-patch usage, support latest contracts, other improvements

shamsartem:


shamsartem:
Merge to enable testing

